PR #6, opened 2019-09-19
Updating Readme

rlancaste:


knro:
Awesome work Robert, this would make it much easier!

PR #10, opened 2019-10-20
Fixing debian files

rkaczorek:
This patch removes various errors to debian files


knro:
Thanks! Though I'm quite surprised as some of these are building already fine on Launchpad?!

rkaczorek:
I don't know why it goes well on launchpad. While packaging locally it didn't work on debian

PR #13, opened 2019-11-02
Splitting indi-asi

rkaczorek:
INDI driver splited to:
- libindi
- indi-asi (linked to libindi)

asi-common is obsolete after this change, becoming part of libasi. Debian files have been updated to recognize new dependencies (indi-asi & indi-ssag -> libasi, libasi conflicts and replaces asi-common).

The PR contains minor changes to other debian packaging files for consistency (typos, incorrect dates).


debian/indi-ssag/changelog
```diff
@@ -1,3 +1,9 @@
+indi-ssag (1.0-1) buster; urgency=medium
+
+  * Changing dependency to libasi
```
knro: Why SSAG depends on asi?

indi-asi/CMakeLists.txt
```diff
@@ -5,25 +5,28 @@ LIST(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules/")
...
-# Using ASI ST4 SDK Version 1.0 updated on 2018-07-23
-# Using ASI EAF SDK Version 0.1.0524 updated on 2019-06-20
 set(ASI_VERSION_MAJOR 1)
-set(ASI_VERSION_MINOR 3)
+set(ASI_VERSION_MINOR 4)
+
+set(ASICAM /usr/lib/arm-linux-gnueabihf/libASICamera2.so)
+set(ASIEFW /usr/lib/arm-linux-gnueabihf/libEFWFilter.so)
+set(ASIST4 /usr/lib/arm-linux-gnueabihf/libUSB2ST4Conv.so)
+set(ASIEAF /usr/lib/arm-linux-gnueabihf/libEAFFocuser.so)
```
knro: Why are these hard-coded to arm?

debian/indi-ssag/changelog
rkaczorek: Originally it depends on asi-common, which will be replaced by libasi. I don't know where the original dependency come from.

indi-asi/CMakeLists.txt
rkaczorek: Good point. How should we pull the libs for various architectures? The present FindASI.cmake pull only lib for CCD. It should be using components to get all the libs. Can you handle it?

knro: Right components is the way to resolve this. I'll see if I have time today for this.

debian/indi-ssag/changelog
knro: It think it was there by mistake, so this dependency needs to be removed.

rkaczorek: Done

indi-asi/CMakeLists.txt
rkaczorek: I implemented it using much simpler approach than components. All libs are returned by findASI.cmake in ASI_LIBRARIES

cmake_modules/FindASI.cmake
```diff
@@ -22,12 +22,36 @@ else (ASI_INCLUDE_DIR AND ASI_LIBRARIES)
...
+  find_library(ASIEAF_LIBRARIES NAMES EAFFocuser
+    PATHS
+    ${_obLinkDir}
+    ${GNUWIN32_DIR}/lib
+  )
+
+
+  if (ASICAM_LIBRARIES AND ASIEFW_LIBRARIES AND ASIST4_LIBRARIES AND ASIEAF_LIBRARIES)
+    set(ASI_LIBRARIES ${ASICAM_LIBRARIES} ${ASIEFW_LIBRARIES} ${ASIST4_LIBRARIES} ${ASIEAF_LIBRARIES})
+  endif (ASICAM_LIBRARIES AND ASIEFW_LIBRARIES AND ASIST4_LIBRARIES AND ASIEAF_LIBRARIES)
```
rkaczorek: This is how FindASI.cmake returns all libs in one variable ASI_LIBRARIES

libasi/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+    exec_program(cp ARGS ${CMAKE_CURRENT_SOURCE_DIR}/x86/libEFWFilter.bin ${CMAKE_BINARY_DIR}/libEFWFilter.so.${ASIEFW_VERSION})
+    exec_program(cp ARGS ${CMAKE_CURRENT_SOURCE_DIR}/x86/libUSB2ST4Conv.bin ${CMAKE_BINARY_DIR}/libUSB2ST4Conv.so.${ASIST4_VERSION})
+    exec_program(cp ARGS ${CMAKE_CURRENT_SOURCE_DIR}/x86/libEAFFocuser.bin ${CMAKE_BINARY_DIR}/libEAFFocuser.so.${ASIEAF_VERSION})
+  ENDIF ()
+ENDIF(UNIX AND NOT WIN32)
+
+install( FILES ${CMAKE_BINARY_DIR}/libASICamera2.so.${ASICAM_VERSION} DESTINATION ${LIB_INSTALL_DIR}${LIB_POSTFIX})
+install( FILES ${CMAKE_BINARY_DIR}/libEFWFilter.so.${ASIEFW_VERSION} DESTINATION ${LIB_INSTALL_DIR}${LIB_POSTFIX})
+install( FILES ${CMAKE_BINARY_DIR}/libUSB2ST4Conv.so.${ASIST4_VERSION} DESTINATION ${LIB_INSTALL_DIR}${LIB_POSTFIX})
+install( FILES ${CMAKE_BINARY_DIR}/libEAFFocuser.so.${ASIEAF_VERSION} DESTINATION ${LIB_INSTALL_DIR}${LIB_POSTFIX})
```
knro: One very important thing that I forgot is that all these files are archive and not shared libraries! So how did you link with them?

rkaczorek: Oh, that's interesting. They link without problem during compilation time and the binary works ok. I have tested it on my ASI camera before PR

rkaczorek: I have just run ldconfig and it returns "is not an ELF file - it has the wrong magic bytes at the start" for each of these files. Also file returns "current ar archive". What can we do about it?

knro: I believe ASI releases both .ar and .so versions.. can you try their latest SDK and see if it resolves this issue?

rkaczorek: I got SDKs and they provide so files indeed. The problem is that the SDK for ST4 improperly linked on ZWO site so I'm missing this library (USB2ST4_linux_mac_SDK)

rkaczorek: This product is discontinued (see https://astronomy-imaging-camera.com/product/zwo-usbst4-guider-adapter) so they don't provide SDK. If you have the SDK for it I can include it, otherwise this lib should be linked the old way in indi-asi :-/

knro: Ok, so maybe stick to static linking for that component?

openastroproject: I believe I have a copy of the SDK including the shared objects, if that would be helpful?

rkaczorek: James, that would be just perfect. Could you drop me an email with it?

rkaczorek: I have everything but USB2ST4_linux_mac_SDK

openastroproject: Should be on its way.

rkaczorek: All done. Thanks for the libs James!

PR #15, opened 2019-11-03
Added Bayer filter support for color cameras

karlrees:
- Added switch to specify whether camera has a Bayer filter or not.
- Enable Bayer-related text fields when Bayer filter selected.
- Update minor version number.

### DIFF
--- a/debian/indi-sx/changelog
+++ b/debian/indi-sx/changelog
@@ -1,3 +1,9 @@
+indi-sx (1.14) bionic; urgency=low
+
+  * Added Bayer filter support option for color cameras
+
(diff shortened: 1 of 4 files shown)


karlrees:
By the way, I was originally hard-coding the driver to automatically recognize color models (those passing the IsColor()) test as having a Bayer filter.  I think this would be more user-friendly for new users, as it avoids the need for the user to learn about and enable the switch.  However, I decided I know too little about the product line and use cases to make such a change at this point, so a switch would be the less disruptive approach.

knro:
Very useful change, thank you! Can you also update the debian changelog as well?

karlrees:
Okay, not entirely sure about the format / conventions of the debian changelog, but hopefully that'll do.

PR #25, opened 2019-11-17
Add files via upload

GiGa-astro:
New meteostation2_sk.xml file in support of the updated meteostation firmware provided previously. If using the bmp085, dht21 and irradiance sensors, the user selects the original meteostation_sk.xml file. If using the bme280 and tsl2591 sensors, this file, meteostation2_sk.xml, should be renamed to meteostation_sk.xml and used before KStars/EKOS/indilib are started.

Thanks for your consideration.

Gilles

### DIFF
--- a/indi-duino/devices/meteostation2_sk.xml
+++ b/indi-duino/devices/meteostation2_sk.xml
@@ -1,0 +1,116 @@
+<INDIDriver>
+
+<defNumberVector device="Arduino MeteoStation" name="Meteo" label="Meteo" group="METEO" state="Idle" perm="ro" timeout="60">
+    <defNumber name="T" label="Celsius" format="%4.2f" min="-100" max="100" step="1">
+    <indiduino pin="24" type="input" mul="0.05" add="-273"/>
+0
+    </defNumber>
+    <defNumber name="HR" label="HR" format="%4.2f" min="0" max="10240" step="1">
+    <indiduino pin="18" type="input" mul="0.01" add="0"/>
+0
+    </defNumber>
+
+    <defNumber name="P" label="mBars" format="%4.2f" min="0" max="10240" step="1">
+    <indiduino pin="16" type="input" mul="0.1" add="0"/>
+0
+    </defNumber>
+    <defNumber name="DEW" label="Dew point" format="%4.2f" min="-100" max="100" step="1">
+    <indiduino pin="20" type="input" mul="0.05" add="-273"/>
+0
+    </defNumber>
+
+    <defNumber name="clouds" label="clouds" format="%4.0f" min="0" max="100" step="1">
+    <indiduino pin="22" type="input" mul="1" add="0"/>
+0
+    </defNumber>
+    <defNumber name="SkyT" label="Sky Temp" format="%4.2f" min="-40" max="40" step="1">
+    <indiduino pin="23" type="input" mul="0.05" add="-273"/>
+0
+    </defNumber>
+
+    <defNumber name="Sky_Quality" label="Sky Qual mag/arcsec2" format="%4.2f" min="0" max="10240" step="1">
+    <indiduino pin="21" type="input" mul="0.01" add="0"/>
+0
+    </defNumber>
+
+</defNumberVector>
+<defLightVector device="Arduino INPUTS" name="STATUS" label="STATUS" group="METEO" state="Idle" perm="ro" rule="AnyOfMany" timeout="60">
+
+    <defLight name="clouds" label="Cloudy">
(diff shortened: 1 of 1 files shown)


knro:
Would there be an easier way to do this without asking the user to rename? how about making symlinks and then based on the binary file name, load the appropriate XML file?

GiGa-astro:
As the firmware name is available to the driver I assume it could be done with simlinks but unfortunately that's a bit out of my league (and comfort zone); not that I could not try but it could take me some time.  I'll take a look at the driver, it may be relatively simple but I will be venturing in not well known territory. Thanks.

knro:
Actually, it might be even simpler. In indi_duino.xml.cmake you can see this:

```
<devGroup group="Weather">
    <device label="Arduino MeteoStation" skel="meteostation_sk.xml">
            <driver name="Arduino MeteoStation">indi_duino</driver>
            <version>@DUINO_VERSION_MAJOR@.@DUINO_VERSION_MINOR@</version>
    </device>
</devGroup>
```

Can't we add another device entry with a different label and skel attribute?

sterne-jaeger:
Gilles, please be aware that we [recently](https://github.com/indilib/indi-3rdparty/commit/249497301006b7c8e25eff017a0782eff5768dc5) changed the parameter names in meteostation_sk.xml. Now all the parameters shown in the group "METEO" are named as WEATHER_... 

This is necessary so that the weather module of EKOS can recognize them.

It would be great if you do it here as well.

Wolfgang

sterne-jaeger:
By the way: shouldn't it be possible using the new firmware with the same skeleton?

GiGa-astro:
To Jasem, after a (not so good) night sleep, I am not sure I know enough on how the drivers and skeleton files 'interact' with each other but I think I start grasping the general concept and I will look into that. I may have further questions though. Thanks.

To Wolfgang, I unfortunately saw the parameter name changes after I submitted my pull request. I will look into changing my submitted skeleton file but, your last comment made me thinking. Before I modified the original skeleton, I used it with the updated firmware and it worked, with the exception of the irradiance/SQM values since the irradiance can range from 0 to 10240 but the SQM will range from small negative values (in bright sunlight) to values around 22-23 (depending on how lucky you are to be in a dark site). I am not sure how we could manage displaying either those two values using the same skeleton file as the irradiance is displayed as-is but the SQM is divided by 100 to display values representative of SQM. 

Thank you both for your insight.

Gilles

GiGa-astro:
I just tested the updated meteostation firmware with the official meteostation_sk.xml file and it works with the exception of one discrepancy, the irradiance vs SQM value display. To display correctly, the  SQM value should be multiplied by 0.01 and use a "%4.2f" format instead of "%5.0f" to account for the decimals, as it is multiplied by 100 in the firmware.

Any suggestions as how this should be dealt with?

I think that we could use the the same scaling (0.01) and format (%4.2f) whether a solar cell (irradiance) or a TSL2591 is used as there are differences in solar cells that make them output different values from one sample to another. The TSL on the other hand will output values that are repeatable (within statistical limits) between samples and are likely more useful as a measure of sky quality.

GiGa-astro:
Latest news/results: As suggested by Jasem, I modified indi-duino.xml to add another meteo station option: Meteo Station SQM, and it loaded the appropriate xml file. Was rather easy, my 'fears' were uncalled for :-). So it seems that there are no real issues anymore to provide the updated firmware to the members of the group. You can disregard the current pull request for the skeleton file and I will provide a new one for the updated indi-duino.xml as well as a new meteostationSQM_sk.xml, before tomorrow end-of-day, hopefully.

Thanks for everything.

sterne-jaeger:
Gilles, it sounds like it would be possible to have one skeleton, if the firmware transforms both TSL2591 and solar cells values to the same value ranges. It would be great if we had one skeleton and one firmware for all sensor types.

GiGa-astro:
Totally agree with you Wolfgang.

Do you see a problem if the TSL2591 value range is used for both, as the SQM value is likely more significant to the end users, and the irradiance value and limits from the irradiance sensor have to be tuned by the user anyway?

sterne-jaeger:
Right, the SQM value is the better one. We need an adaption in meteoWeb, that uses the irradiance value, but that should be straight forward.

GiGa-astro:
I will issue a new pull request with the updated firmware and skeleton files hopefully tomorrow, if that's the proper way to do it.

Thanks.

sterne-jaeger:
Makes sense bundling everything into one PR.

knro:
Ok, will close this PR and the other one and await for a new PR.

PR #26, opened 2019-11-17
Add files via upload

GiGa-astro:
New version of indiduinoMETEO.ino, now supports the BME280 temperature, humidity and pressure sensor that can replace the DHT and BMP sensors; and the TSL2591 light sensor that can replace the irradiance sensor. The TSL sensor provides and estimate of the sky quality. #define are used to specify the connected sensors and check for 'conflicting' sensor defines (e.g. BME280 and BMP085).

Another pull request will provide the xml file which is slightly different than the original one.

Thanks for your consideration.

### DIFF
--- a/indi-duino/devices/Firmwares/indiduinoMETEO.ino
+++ b/indi-duino/devices/Firmwares/indiduinoMETEO.ino
@@ -1,0 +1,1253 @@
+/* INDIDUINOMETEO FIRMWARE.
+  NACHO MAS 2013. http://indiduino.wordpress.com
+  Magnus W. Eriksen 2017 https://github.com/magnue
+
+  Several modifications over indiduinoTemplate:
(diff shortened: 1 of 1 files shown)


sterne-jaeger:
Is there a reason that you placed the firmware in a different location as the old one? Why not simply updating the old indiduinoMETEO.ino?

I checked it locally, everything went well. In addition, I would recommend that we remove the sensor specific files from here. They can be downloaded as libraries, there is no need to replicate them here.

GiGa-astro:
Probably a GitHub beginner lack of knowledge and confidence as to how to proceed. I do not have a problem updating the original file since I had quickly tested it and so did you.

As your recommendation to remove the sensor specific files, I totally agree with you as I find that they clutter the project and they are available as libraries that are usually kept up to date.

PR #28, opened 2019-11-19
Update ASI Camera SDK to version 1.14.1108

jpaana:
Change log:

V1, 14, 1108 2019.11.8
Multi-position for the anti-dew
Add new cameras
Fix some bug


jpaana:
I guess so, added changes to change log

PR #29, opened 2019-11-19
Updated indiduinMETEO.ino and meteostation_sk.xml

GiGa-astro:
Hello Jasem and Wolfgang,

Here are the updated files for the indi-duino meteostation. As I found that the new sensors options created a wrong sensor fail situation, I chased that annoying 'bug' and corrected it (had to add #ifndef verifications for 2 of the sensors).

I am still not sure about the single skeleton file as the irradiance and SQM values do not represent the same thing and have a much different ranges, I'm afraid it may confuse some end users. I like Jasem's idea of having a second virtual version of the driver (i.e. Arduino Meteostation SQM in addition to the original Arduino Meteostation),  that would point via indi-duino.xml to an SQM version of the skeleton file (e.g. meteostationSQM_sk.xml). This is very simply fixed with the following change in indi-duino.xml:

`
<devGroup group="Weather">

    <device label="Arduino MeteoStation" skel="meteostation_sk.xml">
            <driver name="Arduino MeteoStation">indi_duino</driver>
            <version>0.3</version>
    </device>
    <device label="Arduino MeteoStation SQM" skel="meteostationSQM_sk.xml">
            <driver name="Arduino MeteoStation SQM">indi_duino</driver>
            <version>0.3</version>
    </device>
</devGroup>
`
I have a skeleton file I could provide if you want to go that route. I'll leave the decision to you.

Thanks for your insight.


GiGa-astro:
I should have specified that we could put the following, or something similar, in indi_duino.xml.cmake (Jasem's comment from a few days ago) and the required skeleton files would be generated automatically.

```
<devGroup group="Weather">
    <device label="Arduino MeteoStation" skel="meteostation_sk.xml">
            <driver name="Arduino MeteoStation">indi_duino</driver>
            <version>@DUINO_VERSION_MAJOR@.@DUINO_VERSION_MINOR@</version>
    </device>
    <device label="Arduino MeteoStation SQM" skel="meteostationSQM_sk.xml">
            <driver name="Arduino MeteoStation SQM">indi_duino</driver>
            <version>0.3</version>
    </device>
</devGroup>

```

knro:
That looks good. Though we need to use <version>@DUINO_VERSION_MAJOR@.@DUINO_VERSION_MINOR@</version> for both entries. After you installed this to the system, you can run both drivers and they both load the correct skeleton and operate OK?

GiGa-astro:
From what I get, @DUINO_VERSION_MAJOR@.@DUINO_VERSION_MINOR@ are used to identify the indi_duino driver version which I did not touch and is the driver for all the indi_duino devices (Arduino Servo, Arduino Simple Switcher...). So the MAJOR and MINOR version numbers should be the same for all the indi_duino devices.

If I understand correctly, when indi_duino.xml.cmake (or indi_duino.xml) is run, it creates 'virtual' drivers (e.g. Arduino MeteoStation, Arduino Switcher...), that use the indi_duino driver and creates appropriate xml skeleton files for all those devices. In the case of the meteo station, it would create "meteostation_sk.xml" and "meteostationSQM_sk.xml". So, depending on which 'virtual' driver the user selects, legacy or SQM, the appropriate skeleton file would be utilized.

I'll perform some more tests today, thanks for re-building KStars with the latest libraw support :-),  and come back to you with a confirmation if it worked as expected or not. Unfortunately, I won't be able to perform a complete build test as I do not have a test system but I will do a functional test with a modified indi_duino.xml file.

Thanks for your patience.

GiGa-astro:
I am closing this pull request since I submitted a new one that contains all the new and modified files, and has been tested from scratch. Learning is a lifelong process and getting to know better how GitHub works was a humbling one.

PR #30, opened 2019-11-21
Updated Arduino MeteoStation

GiGa-astro:
OK, got it right now, I hope :-). It's a more involved process than I thought and getting older doesn't make it easier but I thought I could contribute so, here it is, sorry for all the trouble and thanks for your comprehension.

All modified and required files should be there. I tested the modifications, even rebuilt INDI from a fresh download, ran KStars/EKOS/INDI and there was no problem. 

The process is simple. After flashing the Arduino meteo station firmware with your specific sensors selection, in INDI you either choose, in the Weather tab of the Profile Editor, "Arduino MeteoStation" or "Arduino MeteoStation SQM" for either the legacy  or SQM-enhanced virtual driver. Once you start INDI and connect to the meteo station, the appropriate skeleton file will be used and the sensor values will be correctly displayed. The MeteoStationWEB should still work with the legacy Arduino MeteoStation driver but I haven't tested it as I do not use it.

### DIFF
--- a/indi-duino/devices/CMakeLists.txt
+++ b/indi-duino/devices/CMakeLists.txt
@@ -6,5 +6,6 @@ install(FILES stepper_sk.xml DESTINATION ${INDI_DATA_DIR})
 install(FILES focuser_sk.xml DESTINATION ${INDI_DATA_DIR})
 install(FILES cosmos_sk.xml DESTINATION ${INDI_DATA_DIR})
 install(FILES meteostation_sk.xml DESTINATION ${INDI_DATA_DIR})
+install(FILES meteostationSQM_sk.xml DESTINATION ${INDI_DATA_DIR})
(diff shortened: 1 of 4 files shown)


knro:
Thanks! Do you think you can update the documentation here? https://indilib.org/develop/arduino/meteostation.html

How about create a driver entry in Auxiliary? https://indilib.org/devices/auxiliary.html

This driver documentation is currently either out of date or severely lacking IMO.

GiGa-astro:
I can take a look at updating the documentation and creating a driver entry, no problem. Is there information/guidelines somewhere on how to proceed to submit new entries?

Thanks!

knro:
Sure that would be great, here are some templates: https://github.com/indilib/indi/tree/master/drivers

There is nothing specific to Auxiliary devices like this, so feel free to adopt it to whatever format you find appropriate.

PR #33, opened 2019-12-03
New Version AOK Skywalker 0.9 with park management

escribana:
[indi-aok.zip](https://github.com/indilib/indi-3rdparty/files/3917584/indi-aok.zip)
This is a version update with minor general corrections and a park management. The driver now implements the possibility to set a park position, to park and unpark the mount.
The start procedure implies now the unparking of the mount as it is intended by INDI for the first action after turning on the controller. To terminate a session the park command will slew the telescope to the predefined position and turn off both Tracking and Mountlock.
The park management was heavily tested and should work well.


knro:
Thank you! Can you please update the Debian change log as well?

PR #40, opened 2019-12-28
[inovasdk] updated i.Nova SDK

iliaplatone:


knro:
Can you please resubmit the library compiled against libc 2.27 or earlier? The updated library was compiled against 2.28 which is not available in 18.04. Even 16.04 had v2.23, so better to compile against that.

PR #41, opened 2019-12-28
[review] Upgrade AAGCloudWatcher to use indiweather

pardo-bsso:
Hi everyone,

this comes from https://github.com/indilib/indi/issues/920

As requested before, it'd be nice if someone else can look at the code, perhaps I made some glaring mistakes but didn't notice.  @FrazCake do you still have access to this device?


knro:
Great so we are finally getting proper INDI::Weather support for this! Awesome. Can you please also submit the debian files so we can build this on the PPA?

pardo-bsso:
Hi @knro  thanks for looking at this.

I added another folder with the debian files but right now I can't make it work. I have INDI installed in a non standard location and the build inside fakeroot fails to find that.

knro:
Perfect, thanks! I'll merge this and asks users to test.

PR #42, opened 2019-12-29
Fix for broken temperature regulation in Indi-qhy

dirkschwarze:
Fix contains 3 changes to file qhy_ccd.cpp:
1. Reduce temp_threshold.
2. Continue to read temperature, even if IPS_OK, to detect drift...
3. Restart regulation if drift causes to leave the threshold area.


knro:
So QHY cameras work in Auto mode and manual mode.. this change caters to both? Manual you just set PWM at a specific value regardless of temperature. While auto you set the temperature and let the driver adjusts the PWM accordingly.

dirkschwarze:
Hello Jasem,
If you press on „cooler off“ in EKOs, the cooler is turned off with PWM going to 0.
In the INDI Control Panel, you can switch to manual and then enter a PWM (Percent) Value.

Directly entering a value, clicking on „set" and it switches to manual automatically does not work, but I do not know if it ever did.


Regards


Dirk





Dirk Schwarze
dschwarze@me.com
mobile: +49 1570 1122 503

knro:
Tested today and at least I don't see regressions with temperature control on QHY163

PR #47, opened 2020-01-11
indi-duino v 0.6 | DHT sensor bug fix & code cleanup

sterne-jaeger:
Changes to indiduinoMETEO firmware
- Support for DHT sensors corrected, now they get initialized correctly.
- Library header files removed - it is much cleaner if these libraries are directly loaded from the Arduino environment.

Changes to Arduino MeteoStation
- The python scripts for accessing the INDI driver of the MeteoStation contains the INDI parameter names hard coded. Now they are configurable from the config file.


knro:
Looks like one temporary file was added to the commit.

sterne-jaeger:
Aehm, which one do you mean? There was a backup file present, which is deleted with this diff.

knro:
Right deleted sorry didn't see that important bit ;)

PR #50, opened 2020-01-18
Update RTLSDR and LIMESDR drivers to spectrograph device

jpaana:
Renamed functions with guidance from radio simulator. Also ran through clang-format as the indentation was a bit all over the place.

### DIFF
--- a/indi-limesdr/indi_limesdr_detector.cpp
+++ b/indi-limesdr/indi_limesdr_detector.cpp
@@ -24,16 +24,18 @@
 #include <indilogger.h>
 #include <memory>
 
-#define min(a,b) \
-  ({ __typeof__ (a) _a = (a); \
-      __typeof__ (b) _b = (b); \
-    _a < _b ? _a : _b; })
-#define MAX_TRIES 20
-#define MAX_DEVICES 4
-#define SUBFRAME_SIZE (16384)
+#define min(a, b)               \
+    ({                          \
+        __typeof__(a) _a = (a); \
+        __typeof__(b) _b = (b); \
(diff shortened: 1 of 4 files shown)


knro:
@iliaplatone Please take a look.

iliaplatone:
That's wonderful!
It misses Spectrum in the sampling rate range, in case the driver is required to do so, I think it's worth to work on the spectrum and histogram helpers. I have to write the DSP plugin for this.

jpaana:
I have the hardware, but haven't really tested it, just wanted to get the driver to compile again :) Is there something I should look into further or are you planning on doing those changes yourself?

iliaplatone:
Your code is the same i would write.
I am guessing if INDI Sensors needs reporting classes or buffer/signal processing methods inside a plugin library.
The data you capture with your driver is the raw data, the spectrum is actually a software hack to obtain such by implementing the helper methods in Spectrograph.
It is good that you removed Spectrum() from there, because I wanted to move Spectrum, Transform and Histogram helpers in the plugin library, as for their INDI properties and tab, so one can choose to use them when needed.
This code should work, I am willing to work on the DSP, if you want you can (should) do the same in LimeSDR (Myriad SDR - to be renamed).
P.S.: I own an RTLSDR but I can't find it right now. If you test it please check out streaming also.

jpaana:
Did the same search & replace exercise to LIMESDR driver, so it now compiles. I don't have the hardware, but hopefully didn't break anything.

knro:
I'll merge this for now to get it compiled.

PR #56, opened 2020-02-06
Replace wrong arm libraries in Intel x86 directory

pchev:
Atik libraries for Intel x86 wrongly install an ARM version making compilation and execution of the i386 atik driver to fail.
~/source/indi/indi-3rdparty/libatik/x86$ file *
libatikcameras.bin: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=23835ac1829d8136c39f67d5cd012ba9adc9e89e, with debug_info, not stripped
libflycapture.bin:  ELF 32-bit LSB shared object, ARM, EABI5 version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=184899e4365d2a46b33bc5b1a310e54e478b68f4, stripped

This commit replace the files by the correct x86 version taken from AtikCamerasSDK-131.zip  in AtikCamerasSDK/lib/linux/x86/WithFlyCapture/

I tested it work with my Atik 314 on Ubuntu i386.


knro:
Thanks, I used the wrong library by mistake.

PR #58, opened 2020-02-16
WIP: Nexstarevo add serial link

fabriziop:
The nexstarevo driver is modified to add the serial link capability to the network link capability already present. Since the serial ports AUX and PC of Celestron scopes requires hardware handshake with echo of transmitted characters, its read/write logic is different from the network link one. I tried to merge them, but, probably, the network link is broken now.

### DIFF
--- a/indi-nexstarevo/NexStarAUXScope.cpp
+++ b/indi-nexstarevo/NexStarAUXScope.cpp
@@ -1,14 +1,16 @@
 
 #include "NexStarAUXScope.h"
 
-#include <inditelescope.h>
-
 #include <algorithm>
 #include <math.h>
 #include <queue>
 #include <string.h>
 #include <unistd.h>
 #include <arpa/inet.h>
+
+#define READ_TIMEOUT 1 		// s
+#define CTS_TIMEOUT 100		// ms
+#define RTS_DELAY 50		// ms
 
 #define BUFFER_SIZE 10240
 int MAX_CMD_LEN = 32;
@@ -212,14 +214,29 @@ NexStarAUXScope::~NexStarAUXScope()
     closeConnection();
 }
 
+const char *NexStarAUXScope::getDefaultName()
+{
+    return (char *)"NexStar Aux Scope";
+}  
+
+const char *NexStarAUXScope::getDeviceName()
+{
+    return (char *)"NexStar Aux Scope";
+}  
+
+bool NexStarAUXScope::ReadScopeStatus()
+{
+    return true;
+}
+
 void NexStarAUXScope::initScope(char const *ip, int port)
 {
     fprintf(stderr, "Preset scope IP %s:%d\n", ip, port);
-    detectScope();
-    bzero(&addr, sizeof(addr));
-    addr.sin_family      = AF_INET;
-    addr.sin_addr.s_addr = inet_addr(ip);
-    addr.sin_port        = htons(port);
+    //detectScope();
+    //bzero(&addr, sizeof(addr));
+    //addr.sin_family      = AF_INET;
+    //addr.sin_addr.s_addr = inet_addr(ip);
+    //addr.sin_port        = htons(port);
     initScope();
 };
 
@@ -233,11 +250,15 @@ void NexStarAUXScope::initScope()
     // Park position at the south horizon.
     Alt = targetAlt = Az = targetAz = 0;
     Lat = Lon = Elv = 0;
-    sock            = 0;
+    //sock            = 0;
+    DEBUG = true;
 };
 
 bool NexStarAUXScope::detectScope()
 {
+    //FP
+    //return true;
+
     struct sockaddr_in myaddr;           /* our address */
     struct sockaddr_in remaddr;          /* remote address */
     socklen_t addrlen = sizeof(remaddr); /* length of addresses */
@@ -296,14 +317,17 @@ bool NexStarAUXScope::detectScope()
     return false;
 }
 
-bool NexStarAUXScope::Connect(int PortFD)
+bool NexStarAUXScope::Connect(int portFD)
 {
+    PortFD = portFD;
+
     if (PortFD > 0)
     {
         // We are connected. Just start processing!
-        fprintf(stderr, "Connection ready. Starting Processing.\n");
-        sock=PortFD;
-        msleep(500);
+        //fprintf(stderr, "Connection ready. Starting Processing.\n");
+	//DEBUG(DBG_NSEVO,"Connection ready. Starting Processing.\n");
+        //FPsock=PortFD;
+        //msleep(500);
         readMsgs();
         return true;
     }
@@ -312,7 +336,9 @@ bool NexStarAUXScope::Connect(int PortFD)
     // This is just to print where is the scope if there is no connection
     if (!detectScope())
     {
-        fprintf(stderr, "Cannot detect the scope!\n");
+        //fprintf(stderr, "Cannot detect the scope!\n");
+	//FP
+        LOG_INFO("Connect: cannot detect the scope!\n");
     }
     return false;
 
@@ -333,12 +359,16 @@ bool NexStarAUXScope::UpdateLocation(double lat, double lon, double elev)
     return true;
 }
 
+// FP modified
 void NexStarAUXScope::closeConnection()
 {
-    if (sock > 0)
+    //if (sock > 0)
+    if (PortFD > 0)
     {
-        close(sock);
-        sock = 0;
+        //close(sock);
+        //sock = 0;
+        close(PortFD);
+        PortFD = 0;
     }
 }
 
@@ -414,6 +444,7 @@ bool NexStarAUXScope::GoToFast(long alt, long az, bool track)
     az %= STEPS_PER_REVOLUTION;
     azmcmd.setPosition(az);
     sendCmd(altcmd);
+    readMsgs();
     sendCmd(azmcmd);
     readMsgs();
     //DEBUG=false;
@@ -442,6 +473,7 @@ bool NexStarAUXScope::GoToSlow(long alt, long az, bool track)
     az %= STEPS_PER_REVOLUTION;
     azmcmd.setPosition(az);
     sendCmd(altcmd);
+    readMsgs();
     sendCmd(azmcmd);
     readMsgs();
     //DEBUG=false;
@@ -466,6 +498,7 @@ bool NexStarAUXScope::Track(long altRate, long azRate)
     azmcmd.setPosition(long(std::abs(AzRate)));
 
     sendCmd(altcmd);
+    readMsgs();
     sendCmd(azmcmd);
     readMsgs();
     return true;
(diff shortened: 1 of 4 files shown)


jochym:
Let me look into it a little @fabriziop. I fully intend to merge it soon but I would like to avoid any regressions in the driver.

fabriziop:
I agree with you. Take your time.
To help to understand my changes, here are some details.

Celestron serial protocol (ports AUX and PC)
1. set RTS on.
2. wait for CTS on with timeout. If timeout abort sequence.
3. send scope command from PC to scope.
4. wait for the physical send completion.
5. set RTS to off. If RTS is set to off before physical send completion of command, the command transmission is truncated.
6. receive echo of scope command. Check for identity with the sent one.
7. receive command response, process it as usual.

For this type of handshake, it is convenient to read command responses after each command sent.
This behavior is different from your net link implementation that uses a pipeline scheme for sending e receiving. This is the most tricky part, I think, the merge of the two communication logic.

F.

fabriziop:
I just pushed a fix. There were a lot of missing calls to readMsgs after sendCmd calls. With serial link logic, readMsgs must be called just after each call to sendCmd. Sorry for my fused brain.

knro:
@jochym @fabriziop is this ready to land?

jochym:
I had no time to work on this. I will test this tomorrow. So hold it for now please @knro .
Sorry @fabriziop - I was busy last few days.

jochym:
@fabriziop : I had time to test on my mount while connected through HC serial socket. 
The connection does not work. If I connect with celestron_gps driver it works fine with the same serial connection and settings. With this version I get CTS timeout right after connection. 
Is it supposed to work over HC bridge, or you need to have hardware interface to AUX port?

fabriziop:
@jochym: this serial link works only with direct connection from serial PC interface of from a USB/serial converter to the AUX or PC ports of Celestron scope. I tested it connected to the PC port of the scope. The HC serial port does not work because it has no hardware handshake, no CTS signal.

jochym:
Why we need CTS/hardware handshake? CTS is technically not required for serial transmission. And as far as I know nexremote (which is effectively the same as our driver) works over this connection. Of course, it is better to have the PC port working alone then no serial connection at all, but maybe we can make the HC port work as well? Many users have no pc-port on the scope (only aux) and you either need expensive celestron accessory or build a replacement yourself. For some people both are not an option. (e.g. I have no real use for such a cable/and no time to build one just for testing - and there is real possibility to brake the scope that way).

fabriziop:
May be I have misunderstood the primary target of nexstarevo driver: for me, direct connection computer<->scope without hc. But never mind, if the real scenario is: I want the driver to work with HC serial and, may be, with AUX or PC serial, some extensions are required.
First the driver must be aware of which serial connection is using (HC or AUX/PC). This can be done manually, setting a property, or better by an automatic test of HC presence. Then the serial read/write functions can be easily modified to work with (AUX/PC serial) or without (HC serial)  RTS/CTS handshake.
More effort is required to adapt the sending of AUX command through the HC serial, since this requires to encapsulate the AUX commands with the passthrough mode, while the PC serial accept the AUX commands as they are.
If you agree with this last driver scenario, I can work on it in the next days.

jochym:
@fabriziop let me state what was my intention at the start of this: write an AUX driver for the scope to have it working with indi the way the eqmod driver does. And yes, using the scope without the HC was a primary goal. In the HC pass-through scenario it just works as a convoluted serial interface. Since there are three channels you can use to send AUX commands my ultimate goal was to have it work over all of them. I started from the network because it was the scope I have (Evo) and it was the driver I personally needed. So my motivation was quite close to yours - we just have different scopes ;). You have no obligation to work on this but if you would care to try, it will be very helpful. At the moment I have no way to test direct connection (there is no PC port in NSE). But I like the factoring out of the communication to standard library functions you did and the implementation itself. So my proposal is: I will make sure the changes have no regressions for network comm and you will work on both channels of serial comm.
Regarding serial: 
1. Are you absolutely positive that PC port comm *requires* CTS/RTS hardware handshake to work? Is it written somewhere? Because it is often looped over in sockets/plugs and the comm is run just over TX/RX lines. 
2. I believe, the encapsulation is quite easy - just slap 0x50 in front of the command (see send_command and send_passthrough in celestrondriver.cpp:147-249). 
We can do it by just switching the prefix on/of depending on connection. 

What do you think?

fabriziop:
@jochym I agree to have all three channels (net, HC, AUX/PC) working as final goal. My current implementation was obviously driven by my setup, a CPC1100 connected via the PC port to an USB/serial converter connected to a raspberry.
You can found a lot of detailed info about nexstar here
https://paquettefamily.ca/nexstar/NexStar_AUX_Commands_10.pdf
See pg. 1 for a general communication scheme and pg. 9 for serial interfacing details.

About your point 1, I can confirm to you that RTS/CTS flow control on CPC PC port is real, I have checked it on my setup. If you do not manage correctly both RTS and CTS, the serial communication does not work. What does not say the documentation above is that the flow control is half duplex. Since there are separated wires for TX and RX, one can think that flow control must be full duplex. That's not the case. I wasted a lot of time to discover it.

About your point 2, ok for the packet format. In addition, It seems to me that passthrough commands are not echoed while the ones on the AUX or PC port are echoed. There is this difference in the communication flow among these interfaces to be managed: echoed chars must be consumed/checked.

So, ok for your proposal, to have a step 1 with both wifi/net and AUX or PC serial port working. I'll wait for your network communication checking.

From my point of view, I will also need to complete the driver by having a mount mode setting (EQN for me) and park position setting. May be these are in your To Do list.

jochym:
Ok. Great. It is most important to know what are the facts "on the ground" - as they say. So there are three modes of communication:
1. Serial over PC-port/adapter:
  - RTS/CTS flow control. 
  - Half duplex
  - Echoed commands
2. Serial over HC (AFAIK)
  - No hardware flow control (four wire connector serial)
  - No command echo
  - Command reply echoed and filtered by HC
  - No broadcast commands. Only command-reply trafic
  - Command prefixing required
3. Network
  - Closer to Serial over Ethernet (p1 above) - the adapter inside is actually serial-over-wifi chip
  - Full AUX bus traffic (all trafic is visible - e.g. emulating GPS accessory or other devices is possible)
  - Obviously no RTS/CTS flow control
  - Command echo present (need to consume/ignore echoed trafic). This should be actually used as a form of ACK. If you got no echo the controller did not get the command.

So we need three modes of communication but the first and third is remarkably close to each other. The only difference is lack of flow control which you already handled. 

The proposed plan:
1. Since only you can test serial over PC-port vs. HC please try to include HC in your code
2. I will check if the changes in comm code did not brake network channel and further unify the net/serial comm.
3. Then we merge
4. Profit ;)

jochym:
@fabriziop I have checked the network connectivity, and unfortunately the changes break it badly. There is a message about incorrect ioctrl for the network connection. It pops up in the box so fast that it almost killed session on my 16GB machine ;). I will try to figure out where is the problem and let you know. But that means we definitely cannot merge yet.

jochym:
I succeeded to get some logs after connecting it goes like this:
```
INFO    82.161885 sec   : Connecting to 10.24.48.10@2000 ...
DEBUG   82.165452 sec   : Connection successful, attempting handshake...
ERROR   82.165758 sec   : Error getting handshake lines Inappropriate ioctl for device(25).
ERROR   82.166049 sec   : Error setting handshake lines Inappropriate ioctl for device(25).
[...]
ERROR   82.169817 sec   : Error setting handshake lines Inappropriate ioctl for device(25).
INFO    83.171213 sec   : NexStar Evolution is online.
DEBUG   83.176786 sec   : Configuration successfully saved for DEVICE_ADDRESS.
DEBUG   83.183126 sec   : Configuration successfully saved for CONNECTION_TYPE.
DEBUG   83.187950 sec   : Configuration successfully saved for CONNECTION_MODE.
ERROR   84.189421 sec   : Error getting handshake lines Inappropriate ioctl for device(25).
ERROR   84.308432 sec   : Error getting handshake lines Inappropriate ioctl for device(25).
ERROR   84.308617 sec   : Error waiting for CTS timeout. modem_ctrl = 4 hex.
ERROR   84.308738 sec   : Error getting handshake lines Inappropriate ioctl for device(25).
[...]
```
Trying to figure out where the network connection gets treated like serial.

fabriziop:
Ok. Good! I'll work to add HC serial.

Bad thing. I hope we can discriminate among serial and network in some way.

jochym:
Working on it. I think we are using the connection plugins incorrectly. I have asked @knro for some examples. I also think we should merge both objects (NexStarAUX and NexStarEvo) into one and, eventually, rename the driver to indi_celestronaux_telescope.

fabriziop:
Well, about software structure, since we want to keep all three communication modes, this driver is more similar to celestrongps.cpp than to eqmod. So, I was thinking why not to stole as much code as possible from celestrongps.cpp and to sustitute celestrondriver.cpp with NexStarAUX, may be all in one object.

jochym:
@fabriziop as one object - yes, that is what I had in mind. As for the similarities, the celestrongps uses HC high level commands and pushes alignment and motor control to the HC. This driver handles alignment, tracking, and communication with motor controllers - so rather like the eqmod. It is not as advanced as eqmod but functionality is similar. The difference is mainly in mount "inteligence" which seems to be slightly higher with celestron (less work with tracking is required). Thus the driver is simpler.

jochym:
I think it is wrong to set:
```
setTelescopeConnection(CONNECTION_TCP|CONNECTION_SERIAL);
```
with two types of connections. But if I put only one (e.g. `CONNECTION_SERIAL`) the driver crashes during initialization:
```
2020-02-28T20:13:57: Driver ./indi_nexstarevo_telescope: Preset scope IP 1.2.3.4:2000
2020-02-28T20:13:57: Driver ./indi_nexstarevo_telescope: NexStarEvo instancing
2020-02-28T20:13:57: Driver ./indi_nexstarevo_telescope: 
2020-02-28T20:13:57: Driver ./indi_nexstarevo_telescope: EnumeratePlugins - found plugin SVD Math Plugin
2020-02-28T20:13:57: Driver ./indi_nexstarevo_telescope: stderr EOF
2
```
I am suspecting we are not using connection plugins correctly.

fabriziop:
From inditelescope.cpp line 617

void Telescope::setTelescopeConnection(const uint8_t &value)
{
    uint8_t mask = CONNECTION_SERIAL | CONNECTION_TCP | CONNECTION_NONE;
    if (value == 0 || (mask & value) == 0)
    {
        DEBUGF(Logger::DBG_ERROR, "Invalid connection mode %d", value);
        return;
    }
    telescopeConnection = value;
}

From inditelescope.h line 871

        uint8_t telescopeConnection = (CONNECTION_SERIAL | CONNECTION_TCP);

So, the call to setTelescopeCommunication with serial and tcp seems to change nothing. It is only a way to make connection capabilities explicit. I checked it by commenting the call, nothing changes for my setup.

Otherwise, it think the problem comes from a parallel inheritance. Both NexStarEvo and NexStarAUXEvo inherit Telescope::. (i added it to NexStarAUXEvo) . I think that this is not correct, since the connection plugins executes inside NexStarEvo. A solution can be to remove that inheritance and to pass to NexStarAUXEvo a pointer to the instance of NexStarEvo. So the call to getActiveConnection is run on the proper instance of Telescope.

jochym:
I think you are right. One more indication that we nedd to merge these two objects. The way it is now is artificial and comes from historic reasons. Let us not patch the problem but properly solve it.

fabriziop:
**Code development continues on PR #61**

PR #64, opened 2020-03-03
Save horizon configuration items.

TallFurryMan:
This is a proposal for saving the configuration items "Abort tracking", "Abort slewing" and "Abort Goto" in the context of horizon management. Saving configuration now includes those items.

As for other drivers, configuration should never be saved when the device is not connected, as undefined properties won't be part of the saved data stream. Properties not saved because they are undefined revert to their default value on next load.

Tested with simulation mode, loading/saving, connecting/disconnecting.


TallFurryMan:
(I'll note that as is, azgtibase.cpp and eqmod.cpp do not build on my side because of missing <cstring> and <memory>, but that's another point)

PR #69, opened 2020-03-12
Weatherradio v 0.8 | Documentation added

sterne-jaeger:
Here comes the first version of the documentation - see doc/Readme-WeatherRadio.md. Additionally, the firmware configuration has been streamlined so that all configuration changes can be done from config.h.


indi-duino/add-on/weatherradio/bin/wr_config.py
```diff
@@ -24,13 +24,13 @@
 INDIDEVICEPORT="/dev/ttyUSB0"
 
 # RRD storage holding time series data
-RRDFILE="./weather.rrd"
+RRDFILE="/usr/local/share/weatherradio/weather.rrd"
```
knro: is there a reason this is /usr/local/share and not /usr/share? The reason is if we make a debian package for this, it should be install to /usr/share

sterne-jaeger: Fine for me. I have my stuff always under /usr/local/... I will correct it.

There is also a bugfix for the firmware, so please wait with merging.

sterne-jaeger:
OK, path changed in the config files and in the documentation.

knro:
Thanks, is this ready to land now?

sterne-jaeger:
Yes, everything is ready from my side.

PR #73, opened 2020-03-18
Weatherradio V 0.9 | TSL237 updates + script robustness

sterne-jaeger:
With this release it should be possible to use TSL237 as alternative to the TSL2591 for SQM measurement.

Additionally, the RRD update scripts are improved so that they do not any longer get stuck when there are connection problems or not all sensors connected.


sterne-jaeger:
Outch, please wait, one of the scripts has a problem.

sterne-jaeger:
OK, now it should be ready.

PR #80, opened 2020-03-30
Fix libricohcamerasdk compile error on aarch64

karlrees:
Removed support for indi-pentax using libricohcamerasdk (PTP mode) on 64-bit ARM, since no binary is provided.


knro:
So I'm having [problems building](https://launchpadlibrarian.net/472603219/buildlog_ubuntu-bionic-amd64.libricohcamerasdk_1.1.0~202004031057~ubuntu18.04.1_BUILDING.txt.gz) this on Launchpad. 

Did you submit your patch to libmtp upstream? I checked their git repo but couldn't find your changes? It would better in the long term to rely on the upstream package. We can add the latest from GIT to the INDI repo (like libgphoto2) so that's not an issue. If the GIT libmtp is OK, then let's use that.

karlrees:
The GIT version of libmtp won't work, unfortunately.  This is a heavily customized version of libmtp that Ricoh used for their driver.  They did include a patch file, but it's relative to libmtp as of 2-3 years ago, and looks like a nightmare to merge now, even if I understood what was going on.  I doubt the libmtp maintainers would want to do a wholesale merge of the changes, and I wouldn't be able to recompile libricohcamerasdk against any changes they insisted on.

(I've been trying to get in touch with Ricoh about this and several other issues.  Seems like they should have just branched from libmtp and given the driver a different name.  So far, no luck).

I don't know much about Launchpad, but just doing a search, it seems like there's a way to ignore or override the dependencies for specific libraries?  https://stackoverflow.com/questions/11238134/dpkg-shlibdeps-error-no-dependency-information-found-for

Again, not sure if that's any help, as this is all very much new to me.

knro:
Ok one solution is to use the forked version in a new project and rename it to something else. maybe libmtpricoh or something like that.. because it needs to be separate from the system libmtp. Then it can be installed as a normal library and libricoh can just depend on it instead of having the binary bundled in.

karlrees:
But the libricohcamerasdk library would still be looking for libmtp and not libmtpricoh, and I don't have the source for libricohcamerasdk to recompile.

Perhaps there is a way to edit the dependency directly in the libricohcamerasdk binary?  I'm not immediately sure how to do that, though, and whether editing the binary would be a problem under the license agreement?

knro:
we can edit the binary to link to the forked version, I've done it for libsbig. Not sure about the license stuff, we can get it done and send them a note. Maybe that will wake them up.

Issue #39, opened 2019-12-22
Support for SNAP port on EQMOD Mounts


setnes:
This has been previously discussed here.  -->  https://github.com/indilib/indi/issues/617

I would love to be able to use the SNAP port on my EQ6-R Pro via EQMOD.  Is there a way to do this yet... even via a simple script?

geehalel:
Hi,
I had already implemented that feature in my own fork. I just remake a fork and put it again in a snapport branch. As you notice, I was using a small script to monitor the opening of a file by the gphoto driver.  I wonder if, snooping a ccd device exposure property, the eqmod driver may toggle a snapport when the exposure property turns idle (the ccd is in exposure). Or maybe there could be other usages...

setnes:
I built the code in your snapport branch, but I don't see any new INDI parameters for the snap ports.  I'm not sure what I should expect to see though.  I have an EQ6-R Pro.

setnes:
I put a "return true;" inside the HasSnapPort1 function and was able to make it work.  I plugged a camera (Fujifilm X-T20) into the snap port and was able to open and close the shutter using the INDI parameter.  So... the next thing you'll want is a mount code.  

`22:43:20.100: [DEBUG] InquireBoardVersion(): MountCode = 35, MCVersion = 30723, setting minperiods Axis1=6 Axis2=6`

Should we assume that the 35 printed in the log should be a 0x23 in the code?  Let me try that.

geehalel:
Thank you for testing, and solving the issue with EQ6-R (I did'nt know the mount code and did not want to look at the firmware). I added the mount in my repo so it should be ok now. I've just tested
with the simulator giving a 0x23 mount code. Maybe could you test again before I make a PR.
Thanks again.

setnes:
There might be two mount codes for EQ6-R Pro.  There are different firmware downloads for mounts with USB and mounts without.  I have a newer one that has USB.  I know someone with the older version of the mount, but we have stay-at-home orders right now for COVID-19.  I'm not sure if I'd be able to step him through finding the mount code remotely.

I will test your new version, but cannot do it until after work.  I also did not try your script (from your older branch) to make this available to gphoto.  Instead I tested by using indi_setprop to open and close the shutter from a simple bash script.

geehalel:
Ok, you're right, there is also a firmware V2.11 for AZEQ6 and EQ6-R (not pro?) mounts on the
skywatcher download page. Let's wait and know more about that later.
[...]

PR #82, opened 2020-04-01
indi-eqmod: Implement Snapport control 

geehalel:
This fixes the issue #39 and besides adds support for EQ6-R Pro mount.

- adds one or two snap port switches in the options tab depending on the mount model (hardcoded in HasSnapPort functions)
- adds EQ6-R Pro mount model with code 0x23 (with a built-in USB port)

### DIFF
--- a/indi-eqmod/eqmodbase.cpp
+++ b/indi-eqmod/eqmodbase.cpp
@@ -360,6 +360,15 @@ void EQMod::ISGetProperties(const char *dev)
             defineSwitch(DEPPECSP);
         }
 
+        if (mount->HasSnapPort1())
+        {
+            defineSwitch(SNAPPORT1SP);
+        }
+        if (mount->HasSnapPort2())
+        {
+            defineSwitch(SNAPPORT2SP);
+        }
+
 #ifdef WITH_ALIGN_GEEHALEL
         if (align)
         {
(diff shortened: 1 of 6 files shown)


ACMACMACM:
> This fixes the issue #39 and besides adds support for EQ6-R Pro mount.
> 
> * adds one or two snap port switches in the options tab depending on the mount model (hardcoded in HasSnapPort functions)
> * adds EQ6-R Pro mount model with code 0x23 (with a built-in USB port)

Hi geehalel, 

I have an EQ6-R Pro mount and would very much like to use the Snap Port.  I can see the option in the Ekos>EQMount>Options menu and can turn it on, but after that, how can i control it?  There doesnt seem to be anywhere ive found to be able to then control it thereafter.

geehalel:
Hi @ACMACMACM 
I'm not sure to understand what you mean by controlling the snap port. As is, the indi-eqmod driver toggles your snap ports when you use the switch in the Options tab. It is the role of a client to decide when to toggle the snap ports. If you want to use it for controlling the shutter release of a DSLR (I do for my old 350D), you may use a small script which will toggle the snap port when gphoto ccd starts an exposure. The code is [here](https://github.com/geehalel/indi/blob/eqsnap/3rdparty/indi-eqmod/eqsnap1.py). Use the file name in this script (<code>/tmp/eqsnapport1</code>) in the 'Shutter Release Port' option in gphoto CCD, and run the script from a terminal (exit with Ctrl-C).

ACMACMACM:
Ok so Ekos would be the client in this case.  Id like to use the Capture module in Ekos to control the snap port of the camera.  Queue up a sequence of exposures in Ekos and instead of GphotoCCD controlling the camera via USB (which it cant do for my camera at this time) it would instead control the camera via the snap port solely.  No USB cable to the camera at all.  I didnt think this would work given Ekos needs to detect a camera in order to start, and to detect a camera it would have to be plugged in via USB (which im trying to avoid).

ACMACMACM:
Thanks, that makes sense that the driver is doing it's part by toggling the snap port, and that a client would have to then control that toggling process. 

Can I ask what to do from here though as i'm not very technical.  Can i queue up exposures (eg. 20x60s exposures with an initial wait time of 5s, with a gap of 3s between each exposure) without the use of Ekos or GPhoto CCD, and only toggle the EQMod Mount snap port via terminal? What might that command line syntax look like?

PR #86, opened 2020-04-04
Changed libmtp to libmtpricoh.

karlrees:


cmake_modules/FindPENTAX.cmake
```diff
@@ -40,13 +40,11 @@ else (PENTAX_INCLUDE_DIR AND PENTAX_LIBRARIES)
         PATHS
         ${_obLinkDir}
         ${GNUWIN32_DIR}/lib
-        PATH_SUFFIXES indipentax
       )
-      find_library(RICOHMTP_LIBRARIES NAMES libmtp.so.9.3.0 
+      find_library(RICOHMTP_LIBRARIES NAMES libmtpricoh.so.9.3.0
```
knro, line 44: I think just mtpricoh should be sufficient here.

PR #93, opened 2020-04-14
Interferometer

iliaplatone:
Added telescope array correlator driver


indi-ahp-interferometer/indi_interferometer.xml.cmake
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<driversList>
+<devGroup group="CCDs">
```
knro, line 3: Should they get their own group? we have receivers IIRC.

iliaplatone: This code refers to https://github.com/iliaplatone/ahp-interferometer
There is both the chance to use it with receivers (ttl antennas) or photomultiplier/APDs. 
Since the output is a picture it's group should be CCDs.

PR #109, opened 2020-04-27
AOK-Skywalker Driver reached the productive status!

escribana:
AOK-Skywalker Driver reached the productive status with version 1.0. It has been tested thoroughly at our observatory for several months now and showed no flaws. So I think it is time for an update.

The updated manual with new snapshots is attached.
[indi_aok.zip.zip](https://github.com/indilib/indi-3rdparty/files/4540190/indi_aok.zip.zip)

(Sorry, that I included some changes in "indi-duino". It happened not on purpose. I attempted to correct some flaws, that showed up in this driver. How can I withdraw this commits from this call? To whom shall I direct the indi-duino request?)


knro:
So I tried to revert and delete here but didn't work. I'll close this and you can submit another PR without the induino commits.

PR #114, opened 2020-04-30
Productive version 1.0 of AOK Skywalker

escribana:
This is the thoroughly tested AOK Skywalker driver at our observatory in the last few months. It behaves quit well, so I think it's time for an update.

0.9 -> 1.0
Cleared up the handling of notification of mountlock & trackstate and
added query of mount when connecting, so mountlock and trackstate is
always reflected correctly on start or reconnect.

There are some changes in the UI as well, so [here](https://github.com/indilib/indi-3rdparty/files/4559845/indi_aok.zip.zip) is the updated manual including the new screenshots.


knro:
Maybe also update the debian files?

escribana:
Are there more debian files than "ChangeLog"?

PR #115, opened 2020-05-01
Arduino driver revamped

escribana:
Correction of a few flaws and addition of a simple "roof" interface.


escribana:
Sorry, but until now I'm not able to separate the two projects with this *** git.

Issue #105, opened 2020-04-26
Atik Horizon 2 not detected


TallFurryMan:
**Describe the bug**
The Atik INDI driver does not detect the CMOS "Horizon II" camera. The default "No Atik cameras detected. Power on?" pop-up appears twice with KStars/Ekos, and the driver is not started in the INDI panel. The unexpected behavior is consistent from the command line.

**To Reproduce**
Exact steps to reproduce the behavior.
1. Connect the Atik Horizon 2 to the Linux computer, in a valid USB2 port.
2. Run `indi_atik_ccd`: `indiserver -l /home/tallfurryman/dev/tmp -p 7626 -vvv indi_atik_ccd`.
3. Message appears in log: `No Atik cameras detected. Power on?`

**Expected behavior**
Atik Horizon 2 is detected and can be used by INDI clients. This same test is able to detect and operate an Atik 314E without issues.

**Screenshots**
Output of `dmesg` relevant for the port the camera is plugged in, showing EHCI:
![image](https://user-images.githubusercontent.com/22483335/80307962-9a4c6f00-87cc-11ea-8478-51f85860ad48.png)
Output of `dmesg` when connecting the camera:
![image](https://user-images.githubusercontent.com/22483335/80307900-548fa680-87cc-11ea-946f-f57fb05886fa.png)
Output of `lsusb` when the camera is connected (Atik Horizon 2 on USB2 port 1, ZWO EFW and Arduino on USB1.1 ports 3 and 4:
![image](https://user-images.githubusercontent.com/22483335/80307917-6ec98480-87cc-11ea-93d1-54f0cffab2e9.png)

**Desktop (please complete the following information):**
 - OS: Lubuntu 18.04 i386
 - Version INDI 3rd-party 7c25c271
 - Atik library in /usr/local/lib/libatikcameras.so.2.2.6

**Log Files**
The only content of the log generated by the test above is the following:
```
2020-04-26T12:51:42: : No Atik cameras detected. Power on?
```
The command line shows several additional lines:
![image](https://user-images.githubusercontent.com/22483335/80308327-ad603e80-87ce-11ea-8af6-708cb9db326d.png)

TallFurryMan:
The first message returned by Atik Support is that the Horizon 2 camera is NOT supported under Linux.

I am trying to understand the situation with them before I am forced to return the camera.

[...]

PR #123, opened 2020-05-07
Update Atik SDK to 20200507-v145.

TallFurryMan:
This PR update the armhf, arm64, x86-linux and x64-linux to version 1.4.0.5, with the relevant `ArtemisCameras.h`.

Files were extracted from https://www.atik-cameras.com/wp-content/uploads/2020/05/AtikCamerasSDK_2020_05_07.zip.

### DIFF
--- a/indi-atik/atik_ccd.cpp
+++ b/indi-atik/atik_ccd.cpp
@@ -2,6 +2,7 @@
  ATIK CCD & Filter Wheel Driver
 
  Copyright (C) 2018 Jasem Mutlaq (mutlaqja@ikarustech.com)
+ Copyright (C) 2020 Eric Dejouhanet (eric.dejouhanet@gmail.com)
 
  This library is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
@@ -29,6 +30,7 @@
 #include <math.h>
 #include <unistd.h>
 
+#define MAX_CONNECTION_RETRIES  5
 #define MAX_EXP_RETRIES         3
 #define VERBOSE_EXPOSURE        3
 #define TEMP_TIMER_MS           1000 /* Temperature polling time (ms) */
@@ -56,34 +58,47 @@ void ATIK_CCD_ISInit()
         iAvailableCamerasCount = 0;
         std::vector<std::string> cameraNames;
 
-        iAvailableCamerasCount = ArtemisDeviceCount();
-        if (iAvailableCamerasCount > MAX_DEVICES)
-            iAvailableCamerasCount = MAX_DEVICES;
-        if (iAvailableCamerasCount <= 0)
-            IDLog("No Atik Cameras detected. Power on?");
-        else
+        IDLog("Atik Cameras API V%d DLL V%d initializing.", ArtemisAPIVersion(), ArtemisDLLVersion());
+
+        for (int loop = 0; loop < MAX_CONNECTION_RETRIES; loop++)
         {
-            for (int i = 0; i < iAvailableCamerasCount; i++)
-            {
-                // We only do cameras in this driver.
-                if (ArtemisDeviceIsPresent(i) == false || ArtemisDeviceIsCamera(i) == false)
-                    continue;
+            iAvailableCamerasCount = ArtemisDeviceCount();
 
-                char pName[MAXINDILABEL] = {0};
-                std::string cameraName;
+            if (0 < iAvailableCamerasCount)
+                break;
 
-                if (ArtemisDeviceName(i, pName) == false)
-                    continue;
+            IDLog("No Atik Cameras detected on attempt %d/%d, retrying...", loop+1, MAX_CONNECTION_RETRIES);
+        }
 
-                if (std::find(cameraNames.begin(), cameraNames.end(), pName) == cameraNames.end())
-                    cameraName = std::string(pName);
-                else
-                    cameraName = std::string(pName) + " " +
-                                 std::to_string(static_cast<int>(std::count(cameraNames.begin(), cameraNames.end(), pName)) + 1);
+        if (iAvailableCamerasCount <= 0)
+        {
+            IDLog("No Atik Cameras were enumerated.");
+            iAvailableCamerasCount = 0;
+            return;
+        }
+        else if (iAvailableCamerasCount > MAX_DEVICES)
+            iAvailableCamerasCount = MAX_DEVICES;
 
-                cameras[i] = new ATIKCCD(cameraName, i);
-                cameraNames.push_back(pName);
-            }
+        for (int i = 0; i < iAvailableCamerasCount; i++)
+        {
+            // We only do cameras in this driver.
+            if (ArtemisDeviceIsPresent(i) == false || ArtemisDeviceIsCamera(i) == false)
+                continue;
+
+            char pName[MAXINDILABEL] = {0};
+            std::string cameraName;
+
+            if (ArtemisDeviceName(i, pName) == false)
+                continue;
+
+            if (std::find(cameraNames.begin(), cameraNames.end(), pName) == cameraNames.end())
+                cameraName = std::string(pName);
+            else
+                cameraName = std::string(pName) + " " +
+                        std::to_string(static_cast<int>(std::count(cameraNames.begin(), cameraNames.end(), pName)) + 1);
+
+            cameras[i] = new ATIKCCD(cameraName, i);
+            cameraNames.push_back(pName);
         }
 
         if (cameraNames.empty())
@@ -103,7 +118,7 @@ void ISGetProperties(const char *dev)
 
     if (iAvailableCamerasCount == 0)
     {
-        IDMessage(nullptr, "No Atik cameras detected. Power on?");
+        IDMessage(nullptr, "No Atik Cameras detected, please connect and/or power on.");
         return;
     }
 
@@ -222,8 +237,9 @@ bool ATIKCCD::initProperties()
 
     // Version information
     IUFillText(&VersionInfoS[VERSION_API], "VERSION_API", "API", std::to_string(ArtemisAPIVersion()).c_str());
+    IUFillText(&VersionInfoS[VERSION_DLL], "VERSION_DLL", "DLL", std::to_string(ArtemisDLLVersion()).c_str());
     IUFillText(&VersionInfoS[VERSION_FIRMWARE], "VERSION_FIRMWARE", "Firmware", "Unknown");
-    IUFillTextVector(&VersionInfoSP, VersionInfoS, 2, getDeviceName(), "VERSION", "Version", INFO_TAB, IP_RO, 60, IPS_IDLE);
+    IUFillTextVector(&VersionInfoSP, VersionInfoS, 3, getDeviceName(), "VERSION", "Version", INFO_TAB, IP_RO, 60, IPS_IDLE);
 
     // Gain/Offset Presets
     IUFillSwitch(&ControlPresetsS[PRESET_CUSTOM], "PRESET_CUSTOM", "Custom", ISS_ON);
--- a/indi-atik/atik_ccd.cpp
+++ b/indi-atik/atik_ccd.cpp
@@ -39,12 +39,12 @@
 
 #define CONTROL_TAB "Controls"
 
-static int iAvailableCamerasCount;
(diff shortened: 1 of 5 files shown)


TallFurryMan:
This PR should not be merged until tested. Additionally, the library version is incorrect. The SDK states `1.4.0.5`, while the `CMakeLists.txt` of the previous `libatik` states a later version.

I need to test my 314E and Horizon II with that new SDK. Other owners of Atik cameras should also support this PR by testing the update with their hardware.

TallFurryMan:
Another feature worth examining in a future PR is support for Fast Mode.

jpaana:
Didn't notice any issues in a quick test with my 383L+

indi-atik/atik_ccd.cpp
```diff
@@ -56,34 +58,47 @@ void ATIK_CCD_ISInit()
         iAvailableCamerasCount = 0;
         std::vector<std::string> cameraNames;
 
-        iAvailableCamerasCount = ArtemisDeviceCount();
-        if (iAvailableCamerasCount > MAX_DEVICES)
-            iAvailableCamerasCount = MAX_DEVICES;
+        IDLog("Atik Cameras API V%d DLL V%d initializing.", ArtemisAPIVersion(), ArtemisDLLVersion());
```
TallFurryMan, line 61: The diff is weird here, but I just added a loop retrying the device detection.
The loop instantiating cameras doesn't count if the enumerated count is zero.

libatik/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 3.0)
 PROJECT(libatik C CXX)
 
-# Built Jan 19th, 2020
-set(ATIK_VERSION "2.2.6")
+# Built May 6th, 2020
+set(ATIK_VERSION "1.4.5")
```
TallFurryMan, line 5: Here I used the same version as the API/DLL version from the SDK. I'm not sure why it was 2.2.6 before, except to match the INDI driver version. But that should be the SDK version actually. Opinion?

rkaczorek:
I have just checked it and the linux & ARM libs in this SDK release are the same as in previous versions i.e. 2.13.3.31. This matches CMakeLists.txt:
`set(FLYCAPTURE_VERSION "2.13.3.31")
set(FLYCAPTURE_SOVERSION "2")
`
2.2.6 is the version of ATIK INDI driver and it should be set to 2.2.7 now, not downgraded to 1.4.5 as you suggest above.

Where does 1.4.0.5 come from. I can't see it anywhere.

libatik/CMakeLists.txt
rkaczorek: This is wrong. Driver's version should be set to 2.2.7

indi-atik/atik_ccd.cpp
```diff
@@ -222,8 +237,9 @@ bool ATIKCCD::initProperties()
 
     // Version information
     IUFillText(&VersionInfoS[VERSION_API], "VERSION_API", "API", std::to_string(ArtemisAPIVersion()).c_str());
+    IUFillText(&VersionInfoS[VERSION_DLL], "VERSION_DLL", "DLL", std::to_string(ArtemisDLLVersion()).c_str());
```
rkaczorek, line 240: What is rationale behind introducing VERSION_DLL? It does not apply to linux, arm and mac libs.

indi-atik/atik_ccd.cpp
```diff
@@ -56,34 +58,47 @@ void ATIK_CCD_ISInit()
...
+        IDLog("Atik Cameras API V%d DLL V%d initializing.", ArtemisAPIVersion(), ArtemisDLLVersion());
+
+        for (int loop = 0; loop < MAX_CONNECTION_RETRIES; loop++)
+        {
+            iAvailableCamerasCount = ArtemisDeviceCount();
+
+            if (0 < iAvailableCamerasCount)
+                break;
+
+            IDLog("No Atik Cameras detected on attempt %d/%d, retrying...", loop+1, MAX_CONNECTION_RETRIES);
```
rkaczorek, line 70: You does not increment loop value in this loop anywhere. It should probably be:
`IDLog("No Atik Cameras detected on attempt %d/%d, retrying...", loop++, MAX_CONNECTION_RETRIES);`

indi-atik/atik_ccd.cpp
```diff
@@ -56,34 +58,47 @@ void ATIK_CCD_ISInit()
...
+        }
+
         if (iAvailableCamerasCount <= 0)
-            IDLog("No Atik Cameras detected. Power on?");
-        else
         {
-            for (int i = 0; i < iAvailableCamerasCount; i++)
```
rkaczorek, line 66: Is this correct? You should be looping among available atik devices not just cameras.
It is always the case that number of atik devices >= numer of atik cameras

TallFurryMan: In the context of indi_atik_ccd, as far as I understood the code, only cameras are filtered in. Filter wheels are managed by the other driver.

indi-atik/atik_ccd.cpp
TallFurryMan: Er, isn't this part of the "for" construct?

rkaczorek: You're right ;-) Sorry about this. I was too fast

rkaczorek:
In general I would suggest to split this pull request as to separate libatik update from indi_atik_ccd update. This way we will make sure that the last working version of indi_atik_ccd driver works with the latest libs. Then we can update indi_atik_ccd driver to use new functions available in SDK.

libatik/CMakeLists.txt
TallFurryMan: Except this is not the driver, this is the SDK library bundle. The driver version is in `indi-atik/CMakeLists.txt`. Unless there is a good reason that I don't know about, this should reflect the API/DLL version, which is returned as `145` by the Artemis API.

indi-atik/atik_ccd.cpp
TallFurryMan: Well, I'd agree if there was a document stating so. You had this information in the past? I couldn't find it in the SDK docs.

indi-atik/atik_ccd.cpp
rkaczorek: Now I see it. It is iAvailableCamerasCount variable name that is confusing. It is assigned value from ArtemisDeviceCount() so in fact it should be called iAvailableDevicesCount.

TallFurryMan: I'll change this, good point.

TallFurryMan:
I added the changes in `indi_atik_ccd` because I observed that sometimes the Horizon 2 would not be enumerated at first attempt. Is it a reason good enough? Agreed, I did not add any new feature, the "Fast Mode" will be added in a later PR.

TallFurryMan:
> Where does 1.4.0.5 come from. I can't see it anywhere.

The Windows libraries do report that version. For Linux, only the API reports this version. Those version matched during my tests on the three platforms.

libatik/CMakeLists.txt
rkaczorek: I think it is legacy versioning. All the other libs use SDK version number. So in general you are right. However as for now it is aligned with version in debian packaging system. If we just decrease version, users will not get the updated libatik (1.4.5 << 2.2.7). Therefore my suggestion is to release the libatik as 2.2.7 and set(ATIK_VERSION "2.13.3.31") based on  libflycapture.so.2.13.3.31. This will be consistent how we version other libs.

rkaczorek:
> > Where does 1.4.0.5 come from. I can't see it anywhere.
> 
> The Windows libraries do report that version. For Linux, only the API reports this version. Those versions matched during my tests on the three platforms.

So there's a mess with versioning then. See my other comment above

indi-atik/atik_ccd.cpp
TallFurryMan: Actually, I think the driver has a bug here: the number of cameras in `cameraNames` does not include the other devices, and connecting for instance a camera and a filter wheel will cause an index overflow...

indi-atik/atik_ccd.cpp
```diff
@@ -56,34 +58,47 @@ void ATIK_CCD_ISInit()
...
+        else if (iAvailableCamerasCount > MAX_DEVICES)
+            iAvailableCamerasCount = MAX_DEVICES;
 
-                char pName[MAXINDILABEL] = {0};
-                std::string cameraName;
+        for (int i = 0; i < iAvailableCamerasCount; i++)
+        {
+            // We only do cameras in this driver.
+            if (ArtemisDeviceIsPresent(i) == false || ArtemisDeviceIsCamera(i) == false)
+                continue;
```
TallFurryMan, line 86: This is causing the index overflow in `cameras`.

indi-atik/atik_ccd.cpp
rkaczorek: So it returns VERSION_DLL = 1.4.0.5 for linux, arm and mac libs? It could be the case that it actually returns windows lib (dll) version instead. I do not recall anything like version 1.4.0.5 in the past.

indi-atik/atik_ccd.cpp
rkaczorek: I use both Atik CCD and EFW and have not observed this in the past.

indi-atik/atik_ccd.cpp
TallFurryMan: I'll keep 1.4.0.5 as version because that matches the SDK versions properly, see investigation in issue #105.
I have no information on what to use between API and DLL, so I can remove my change here, OK.

indi-atik/atik_ccd.cpp
TallFurryMan: If you connect a CCD and an EFW, `iAvailableCamerasCount` will equal 2, but only one of `cameras` cells will be uninitialized (depending on which device is enumerated first), and `cameraNames` will contain only one element. I don't really understand why the loops below are not crashing when de-referencing `cameras` content. Unless I'm missing something?

indi-atik/atik_ccd.cpp
rkaczorek: See my comment above to libatik/CMakeLists.txt

indi-atik/atik_ccd.cpp
TallFurryMan: I chose to add safeguards on the `cameras` element considered.

libatik/CMakeLists.txt
TallFurryMan: Done.

knro:
Did you test this on armhf? does it operate the camera OK?

TallFurryMan:
Tested OK with:
- Horizon 2 on `x64 Linux` and `x86 Linux`
- 314E on `x86 Linux`

I cannot test my own PR on `armhf` because I restrict the non-development StellarMate unit to official updates. I will need the ppa build.

knro:
@rkaczorek Can you please test and confirm it's OK on your end as well?

rkaczorek:
> @rkaczorek Can you please test and confirm it's OK on your end as well?

I will do that

TallFurryMan:
8h of calibration frames captured OK with the Horizon 2. Quick check on frames doesn't reveal any issues.

As a minor point, the driver arch does not allow to connect a new device on-the-fly. The remote server needs to be terminated and restarted for enumeration to proceed again. Not a regression.

knro:
@rkaczorek Any update?

rkaczorek:
@knro Tested and confirmed it's OK. Sorry for delay. @TallFurryMan Well done!

Two remarks:
- debian packaging files for libatik and indi-atik need to be updated
- libatik needs some re-thinking on future versioning because now it's a mess as noted in the discussion over this PR - now it's 2.2.7 which is consistent with incremented value of deb package, however SDK API returns version 1.4.5 (which we cannot use to be consistent with deb package versioning)... in the same time flycapture lib is 2.13.3.31 :-/

knro:
Awesome, thank you for the the feedback. @TallFurryMan Can you please update the debian files as well?

TallFurryMan:
Thanks for merging.
The week was full, so there was no time to check the debian files. I'll see what I can do this week-end.

PR #139, opened 2020-05-25
Updated ASICAM SDK to version 1.5.0430.

wvreeven:
I updated the ASI Camera SDK to version 1.5.0430. Judging from previous changes made by Jarno Paananen and from local tests done on a Ubuntu mini-pc with my ASI6200 camera the changes are correct.


libasi/99-asi.rules
```diff
@@ -1,3 +1,3 @@
-ACTION=="add", ATTR{idVendor}=="03c3", RUN+="/bin/sh -c '/bin/echo 256 >/sys/module/usbcore/parameters/usbfs_memory_mb'"
-# All ASI Cameras and filter wheels
-SUBSYSTEMS=="usb", ATTR{idVendor}=="03c3", MODE="0666" 
+ACTION=="add", ATTR{idVendor}=="03c3", RUN+="/bin/sh -c '/bin/echo 200 >/sys/module/usbcore/parameters/usbfs_memory_mb'"
```
knro, line 1: Any reason for this?

wvreeven: That's what was in the asi.rules file in the SDK. I did this because the allocated USB memory is less and I figured that ZWO have a reason for it. I can revert it if you want.

jpaana:
> Please use dos2unix on the header file. @jpaana Everything looks OK from your side?

Looks good to me apart from debian/libasi/rules file needing the same version number change so that packaging gets correct symlink:

# ASICAM version
ASICAM_VERSION=1.14.1227

wvreeven:
Thanks for the thorough reviews. I ran dos2unix and corrected the debian/libasi/rules file.

libasi/99-asi.rules
knro: Right, it should remain 256 to be consistent at least with the rest of INDI rule files.

wvreeven: OK, done.

PR #143, opened 2020-06-04
BlackLevel and continuous gain scale for LCG/HCG

SevenW:
To improve support for Altair Hypercams I have implemented BlackLevel and aligned the handling of Low and High conversion gain consistent with the way other camera manufactures like ZWO do this.

HCG adds an additional gain to the readout. In the case of an IMX294 this is a factor 4.5. The common approach is to have one continuous gain scale of which the upper limit is the in the case of the example 4.5 times higher than the native sensor upper limit. The switch over happens at the HCG threshold and the driver now calculates the sensor gain to be applied. When the LCG/HCG ratio parameter is changed it is now required to update the range of the gain parameter. [This](https://nighttime-imaging.eu/docs/develop/site/troubleshooting/altair_dualgain/) link provides some more background information.

BlackLevel was not present yet. The BlackBalance is applicable to RGB mode, and BlackLevel to RAW images. In the implemenation a range maximum is set based on the maximum bit depth. This calculation is correct, but a sligth deviation form the toupcam api and could be replaced with some more elaborate code using a switch and the separate defines for the blacklevel maximum values.

### DIFF
--- a/indi-toupbase/indi_toupbase.cpp
+++ b/indi-toupbase/indi_toupbase.cpp
@@ -252,18 +252,26 @@ bool ToupBase::initProperties()
     IUFillNumber(&ControlN[TC_GAMMA], "Gamma", "Gamma", "%.f", 20, 180, 10, 100);
     IUFillNumber(&ControlN[TC_SPEED], "Speed", "Speed", "%.f", 0, 10, 1, 0);
     IUFillNumber(&ControlN[TC_FRAMERATE_LIMIT], "FPS Limit", "FPS Limit", "%.f", 0, 63, 1, 0);
-    IUFillNumber(&ControlN[TC_HCG_THRESHOLD], "HCG Threshold", "HCG Threshold", "%.f", 0, 1000, 100, 0);
-    IUFillNumberVector(&ControlNP, ControlN, 9, getDeviceName(), "CCD_CONTROLS", "Controls", CONTROL_TAB, IP_RW, 60,
+    IUFillNumber(&ControlN[TC_HCG_THRESHOLD], "HCG Threshold", "HCG Threshold", "%.f", 0, 1000, 100, 900);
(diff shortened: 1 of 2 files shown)


indi-toupbase/indi_toupbase.cpp
```diff
@@ -1059,6 +1096,7 @@ bool ToupBase::ISNewNumber(const char *dev, const char *name, double values[], c
                         break;
 
                     case TC_HCG_THRESHOLD:
+                        //TODO: changing the threshold may change the actual gain and CG setting
```
SevenW, line 1099: Comment no longer applicable. To be removed.

knro:
Thank, this would probably also require an update on the documentation in INDI. If everything is left as default, the camera would not apply any Gain Conversion changes to the readout, correct?

SevenW:
Yes indeed.

I could move the two HCG parameters out of ControlNP and put them in a group of their own and display them only when camera supports it.

Let me know.

If you give me appointee to how to update the doc I can write something on HCG.



Verstuurd vanaf mijn iPhone

knro:
Yeah maybe the two HCG parameters can be in their own property that would work as well. For the documentation changes, you can do it here and I'll apply it to INDI documentation page, or if you have INDI account, I can grant you privilege to edit that page.

SevenW:
Now with nice separation of Dual Gain in the control panel. The parameters are updating nicely. I changed the way HCG can be enabled and disabled. This is no longer by the threshold having value zero, but by the LCG/HCG gain ratio being set to 1. I did so because the Threshold should have a lower limit of the gain itself (100 for the Altair 294).

I can update the docs as well. I prefer to have access on the indilib website so I can see what I write. Mu user name is SevenW, so please grant me write access.

knro:
@SevenW Thank you so much! I've granted you access to modify the documentation in the website. It's greatly appreciated.

SevenW:
I have updated the manual of ToupCam and Altair with a text on the dual gain. I can add a screenshot later on. I need to change language somewhere first.

PR #147, opened 2020-06-07
Exclude libricohcamerasdk dependency from arm64 builds

karlrees:
Guess aarch64 was the wrong keyword for this?  Or it could be I'm doing it wrong...


knro:
yes, arm64 is used.

PR #149, opened 2020-06-11
Housekeeping - debian changelogs update

rkaczorek:
Housekeeping


rkaczorek:
I've been trying to keep debian packaging files clean and concise. Apparently they are kept forgotten by devs when they submit PR. As the results they need to be updated at PR approval, or are left outdated. We need a global solution for this.
@knro why don't we move all debian directories to where they belong i.e. to drivers directories. Now they are kept centrally in a single folder. Only packagers know that ;-)

knro:
So the reason they are placed here is due to launchpad limitations back in the day. I guess each could be moved to its own directory. Let me see if this is no longer an issue with launchpad.

PR #152, opened 2020-06-18
Possible fix for buffer saving problems on Pentax K30 and others.

karlrees:
Also added extra debug statements.


knro:
No need to put newline to LOGXXX functions. Also, do you really need IDLog?

karlrees:
Probably not.  I think I used it at first just because I was copying and pasting from another driver that used it.  It did come in handy figuring out some timing issues early on, but it could probably be taken out now.

PR #166, opened 2020-07-15
Update Atik SDK to v2020.06.23.

TallFurryMan:
This PR updates the Atik SDK to version 2020.06.23 (Atik now uses the release date as version).

Changelog: https://www.atik-cameras.com/change-log/
- Fixed the Horizon dark banding issue.

Note the last argument of `ArtemisCameraSpecificOptionsGetData` is now a reference instead of a pointer.

### DIFF
--- a/debian/libatik/changelog
+++ b/debian/libatik/changelog
@@ -1,3 +1,9 @@
+libatik (2.2.8) bionic; urgency=medium
+
+  * Upstream release.
+
+ -- Eric <tallfurryman@Mundiibuntu>  Mon, 20 Jul 2020 21:45:23 +0200
+
 libatik (2.2.7) bionic; urgency=medium
 
   * Upstream release.
--- a/debian/libatik/copyright
+++ b/debian/libatik/copyright
@@ -3,7 +3,7 @@ Fri, 21 Dec 2007 02:44:13 +0300.
 
 Copyright: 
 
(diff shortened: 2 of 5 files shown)


TallFurryMan:
The INDI driver now outputs the following instead of `V145`:
```
Atik Cameras API V20200618 DLL V20200618 initializing.
```

Perhaps I should use that release date in `CMakeLists.txt` instead of 2020/06/23.

knro:
@TallFurryMan all testing went OK I presume? no linking issues?

TallFurryMan:
No linking issue. Quick tests with the Horizon2 yesterday night were OK. I observed that some settings were reset in the camera. 16MP frame takes 40 seconds to download on USB2 on my laptop, against 7 seconds on my minipc with the previous release, so I will need to double check that part. Dark banding seems drastically reduced, but that was not a sky test, so, still need to validate this properly.

Atik Support said they would come back on the reference subject. This point is not blocking though.

rkaczorek:
> I shifted the version to 2.2.8. @rkaczorek, your opinion?

Sure, let's keep it like that.
Please remember to update debian/libatik/changelog. Use `dch -i` to increase version and `dch -r` to mark as released, when ready for release.

knro:
@TallFurryMan Can you update the debian files as well?

TallFurryMan:
Yes I will.

PR #170, opened 2020-07-23
Support for Atik CMOS camera features

TallFurryMan:
This PR brings support for the following features of Atik CMOS cameras:

- Even Illumination: clears the sensor buffers before exposing, avoids gradient artifacts at the expense of some read noise.
- Exposure Speed: allows choosing between PowerSave (lowest noise) and Normal buffer reading.
- Gain/Offset Presets: allows choosing between Custom, LowGain, MediumGain and HighGain gain/offset presets.
- Pad Data: shift pixel values by 4 bits so that the full dynamic range spans [0,65535] instead of [0,4095].

Note that some properties are stored in the camera itself, and will still be configured after a hardware power cycle. There is no known way to reset the hardware to defaults.

You would want Even Illumination enabled to avoid gradients, especially when doing short exposures (flats for instance).

You would want Exposure Speed set to PowerSave for normal low-noise usage. Note that Fast Mode is not implemented yet in this version. This feature will bring streaming support.

You would want Pad Data enabled for your frame calibration ADUs to match a percentage of the full dynamic range, instead of hitting the high limit around 6% (4096).

**Note about Gain/Offset**: For usability reasons, setting Custom Gain/Offset properties will automatically switch the driver to use the Gain/Offset Preset "Custom". In the same logic, the Gain/Offset Preset property will cause the values of the Custom Gain/Offset properties to be read from the camera and updated. However, only the Gain/Offset Preset "Custom" will cause Custom Gain/Offset properties to be loaded from a configuration file. In other words preset "Custom" writes "Gain/Offset" and any other preset value reads "Gain/Offset". Remember this when setting your capture properties.

**Change applying to both CMOS and CCD**: there is no way to obtain the actual possible exposure range from the SDK. This PR adjusts the default range so that all Atik cameras may be used with their exposure specifications, at the expense of failures when trying to use an unsupported exposure duration. Based on official camera specifications, now CMOS cameras get 18us minimal exposure, CCD cameras get 1ms minimal exposure, and all cameras get 24-hour maximum exposure. Please check your hardware documentation for the available exposure range.

### DIFF
--- a/indi-atik/CMakeLists.txt
+++ b/indi-atik/CMakeLists.txt
@@ -14,7 +14,7 @@ find_package(ATIK REQUIRED)
 FIND_LIBRARY(M_LIB m)
 
 set(ATIK_VERSION_MAJOR 2)
-set(ATIK_VERSION_MINOR 6)
+set(ATIK_VERSION_MINOR 7)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/config.h.cmake ${CMAKE_CURRENT_BINARY_DIR}/config.h )
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/indi_atik.xml.cmake ${CMAKE_CURRENT_BINARY_DIR}/indi_atik.xml)
--- a/indi-atik/atik_ccd.cpp
+++ b/indi-atik/atik_ccd.cpp
@@ -296,7 +296,9 @@ bool ATIKCCD::initProperties()
     IUFillSwitch(&FastModeS[FASTMODE_POWERSAVE], "CONTROL_POWERSAVE", "Powersave / Low noise", ISS_ON);
     IUFillSwitch(&FastModeS[FASTMODE_NORMAL], "CONTROL_NORMAL", "Normal", ISS_OFF);
     IUFillSwitch(&FastModeS[FASTMODE_FAST], "CONTROL_FAST", "Fast / Stream", ISS_OFF);
-    IUFillSwitchVector(&FastModeSP, FastModeS, 3, getDeviceName(), "CCD_FAST_MODE", "Fast Mode", CONTROLS_TAB, IP_RW,
+    // TODO: Implement fast mode 'fast' then increase SwitchVector size to 3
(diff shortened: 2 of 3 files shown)


TallFurryMan:
Current test state for this MR is: **fully tested with Horizon2, not tested with 314E**.
Environment: Lubuntu 18.04, i386 Atom with 1GB ram.

The PR impacts CMOS cameras (Horizon, Horizon2 and ACIS7).
There is no expected change for CCD cameras, except the exposure range which is increased.

@knro I can't add the "enhancement" label here.
@rkaczorek @jpaana @pchev, if you have time, I would appreciate your opinion as Atik owners.
I posted the update to https://indilib.org/forum/astroberry/6890-re-astroberry-and-atik-horizon-ii.html to catch the interest of the only other Horizon2 owner I know about.

pchev:
I tested your branch with my 314L+. Work fine, no regression.

jpaana:
I have my 383L+ at home currently and not at the observatory, but I'll try to do a quick dry run during the weekend

jpaana:
Finally had some time to hook up my 383L+ and everything seems to work. With this camera 1ms exposures don't really make sense as it has very slow mechanical shutter and the minimum was 0.1s before and even then suffered from severe shutter shadowing. So that might confuse some people if they try to take too short flats or something, but other than that no complaints.

TallFurryMan:
@jpaana that's true, the minimum exposure spec for the 383L+ is 200ms. I got in touch with Support to determine the exposure interval using the SDK. There is actually no function, and they suggested to set 1ms as a minimum. That is what I settled for for CCDs. What happens if you ask for a 1ms exposure with the 383L+? You get an simple error, or instability?

jpaana:
Nothing bad happens, the shutter just doesn't have time to open so it's basically like a dark frame. Up side is that I can now take shorter bias frames than before (not that there is any difference :)

knro:
Is this good to go now?

TallFurryMan:
My tests with the 314E are OK. I think we're good yes.

With the Horizon2, the GO Controls are usable along with the Presets low/med/high, through the capture properties.

PR #174, opened 2020-07-29
updated for new libraw data structure

### LINKED ISSUE #173: indi_gphoto does not compile with libraw-0.20
Compiling indi-gphoto with libraw-0.20 throws an error because the data structures have changed:

```
.../indi-3rdparty/indi-gphoto/gphoto_driver.cpp:870:39: error: 'struct libraw_imgother_t' has no member named 'SensorTemperature'
  870 |             if (lib_raw.imgdata.other.SensorTemperature > -273.15f)
      |                                       ^~~~~~~~~~~~~~~~~
.../indi-3rdparty/indi-gphoto/gphoto_driver.cpp:871:66: error: 'struct libraw_imgother_t' has no member named 'SensorTemperature'
  871 |                 gphoto->last_sensor_temp = lib_raw.imgdata.other.SensorTemperature;
      |                                                                  ^~~~~~~~~~~~~~~~~
.../indi-3rdparty/indi-gphoto/gphoto_driver.cpp:872:44: error: 'struct libraw_imgother_t' has no member named 'CameraTemperature'
  872 |             else if (lib_raw.imgdata.other.CameraTemperature > -273.15f)
      |                                            ^~~~~~~~~~~~~~~~~
.../indi-3rdparty/indi-gphoto/gphoto_driver.cpp:873:66: error: 'struct libraw_imgother_t' has no member named 'CameraTemperature'
  873 |                 gphoto->last_sensor_temp = lib_raw.imgdata.other.CameraTemperature;
```

By changing `lib_raw.imgdata.other.SensorTemperature` to `lib_raw.imgdata.makernotes.common.SensorTemperature` it compiles again, but I still have to test if everything works. If you want I could open a pull request with the changes.

knallio:
Would fix #173, but breaks compatibility with libraw<0.20. Since libraw-0.20 is rather new it might make sense to wait before merging or to try to support both. Not sure how complicated that would be.


knro:
Perhaps add #ifdef version check? this would make it compile with 0.2 but not with previous versions.

knallio:
I thought about #ifdef checks, but I don't have much experience with them, especially in combination with cmake.

I tried to update the pull request accordingly. But I have the feeling that there should be a better way to do it, maybe with some cmake-magic... Did I mention that I don't have much experience with cmake?

At least at the moment it works as I expected, it compiles with both libraw versions.

knro:
That's actually a very good patch for someone who doesn't know much about cmake! I was thinking more along the lines of libraw version check? FindLibraw.cmake defined LibRaw_VERSION_STRING

perhaps this can be used?

knallio:
I tried the version check first, but I found no consistent way to get the version. FindLibraw.cmake might work, but as far as I can see it is not always present since it is not directly provided by libraw. At least I (running arch Linux) don't have it. In principle cmake knows about the version, I guess from pkgconfig files. But as far as I understood find_package only sets version variables if a specific version is requested. So while cmake knows about the version, I found no direct way to use it.

knro:
Sounds good. One small issue, LIBRAW_CAMERA_TEMPERATURE2 and LIBRAW_SENSOR_TEMPERATURE2 are not defined in config.h.cmake

knallio:
Thanks. I have added the variables.

Issue #180, opened 2020-08-03
Queue Report pile up in indi_gpsd


rtlprmft:
I am not sure whether this is a general issue or depends on my configuration. I am using a ublox7-usb module with indi_gpsd. When I enable the automatic refresh (1s, 5s, or even manually), for every refresh, the time only increases by only one second. To my understanding, this is because the gpsd delivers a continuous stream and buffers this stream. When the buffer is then read with gps->read() only the next (but not the most recent report gets read. Therefore, I suggest the following change (works for me flewlessly):

Wrap the gps->read()  in w a while loop which makes sure that the queue is empty:

    while (1)
    {
        if ((gpsData = gps->read()) == nullptr)
        {
            LOG_ERROR("GPSD read error.");
            IDSetText(&GPSstatusTP, nullptr);
            return IPS_ALERT;
        }

        if (!gps->waiting(0))
            break;
    }

knro:
Thank you, I noticed this issue too. 

@rkaczorek What do you think?

rkaczorek:
Seems reasonable to me. @rtlprmft could you submit PR so I can review complete change?

jochym:
That is an obvious bug. I second the need for a fix. If you have a a patch for that @rtlprmft please submit a PR, otherwise we (authors of gpsd) need to fix this.

rtlprmft:
I am sorry, but I have no idea how to do so. Code snippet that solves the problem (empties the queue) is above. I am happy for any advice.

rkaczorek:
@jochym are you going to handle it?

jochym:
@rkaczorek I'll do it today

PR #182, opened 2020-08-05
Fix data pile-up in gpsd

jochym:
Empty the buffer of gpsd daemon and keep just last segment of data. Fix suggested by @rtlprmft in https://github.com/indilib/indi-3rdparty/issues/180

### DIFF
--- a/indi-gpsd/gps_driver.cpp
+++ b/indi-gpsd/gps_driver.cpp
@@ -241,12 +241,19 @@ IPState GPSD::updateGPS()
         }
         return IPS_BUSY;
     }
-
-    if ((gpsData = gps->read()) == nullptr)
-    {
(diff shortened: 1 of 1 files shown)


knro:
Excellent. I guess there is no way for the while loop to get stuck under any circumstance?

jochym:
That is my worry as well. On the face of it, it should always either get error or empty the buffer. But in reality it may depend on the details of implementation of libgpsd and changes in the API. If you prefer I will implement a timeout mechanism with a message to contact developers if it ever happens. What do you think?

knro:
Yeah I think that's a good idea, it doesn't hurt.

jochym:
Fingers crossed ;)

PR #185, opened 2020-08-07
Adding Simulation Capabilities to gpsd

rtlprmft:
This allows to have coordinates without exchanging the driver even though the GPS is not connected


knro:
I see some comments removed, is there a reason?

jochym:
No. I'll fix this.

Issue #187, opened 2020-08-09
indi_gpsd incompatible with gpsd 3.21


mobarre:
Here is the build error:
```
[ 84%] Building CXX object indi-gpsd/CMakeFiles/indi_gpsd.dir/gps_driver.cpp.o
/home/mobarre/AUR/libindi_3rdparty/src/indi-3rdparty-1.8.5/indi-gpsd/gps_driver.cpp: In member function ‘virtual IPState GPSD::updateGPS()’:
/home/mobarre/AUR/libindi_3rdparty/src/indi-3rdparty-1.8.5/indi-gpsd/gps_driver.cpp:252:18: error: ‘struct gps_data_t’ has no member named ‘status’
  252 |     if (gpsData->status == STATUS_NO_FIX)
      |                  ^~~~~~
make[2]: *** [indi-gpsd/CMakeFiles/indi_gpsd.dir/build.make:82: indi-gpsd/CMakeFiles/indi_gpsd.dir/gps_driver.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2131: indi-gpsd/CMakeFiles/indi_gpsd.dir/all] Error 2
make: *** [Makefile:171: all] Error 2
``
The following comment has been left in gps.h:
      Move gps_data_t->status to gps_fix_t.status for better fix merging
       Add wspeedt, wspeedr, wanglem, wanglet, wangler to  gps_fix_t
       Remove unused gps_data_t.navadata_t.
       Add relPosL and relPosH to gps_fix_t.NED

Looking at their gitlab repo, it seems to have happened with release 3.21 of gpsd

Cheers !

**Comment by knro** [comment]
Any #ifdef patch to fix this by checking the version?

**Comment by knro** [comment]
@jochym What do you think?

**Comment by mobarre** [comment]
#ifdef might work indeed !
in gps.h (from gpsd package) we have:
```
#define GPSD_API_MAJOR_VERSION  10      /* bump on incompatible changes */
#define GPSD_API_MINOR_VERSION  0       /* bump on compatible changes */
```

version 9.1 should have the former structure. 10.0 is the new one.

knro:
Great, can you please submit a PR to check for this?

mobarre:
Sure, with pleasure ! Let me get my shit together, I might be able to send it today

jochym:
[...]

PR #188, opened 2020-08-09
Add conditionnals to detect gpsd API version

mobarre:
Attempted quick fix for issue #187. This will certainly require some testing (don't have a laptop GPS yet, so can't really test this).

### DIFF
--- a/indi-gpsd/gps_driver.cpp
+++ b/indi-gpsd/gps_driver.cpp
@@ -214,6 +214,9 @@ IPState GPSD::updateGPS()
         IDSetSwitch(&RefreshSP, nullptr);
     }
 
+#if GPSD_API_MAJOR_VERSION >= 10
+    struct gps_fix_t *gpsFix;
+#endif
     struct gps_data_t *gpsData;
     time_t raw_time;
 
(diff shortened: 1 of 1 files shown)


jochym:
Why we cannot make just the definition conditional? In the second change we read the same attribute and compare it with the same constant. I am sure you did it for some reason, I just would like to know that reason. ;)

mobarre:
No, I have not done this for a reason, and the suggestions are very welcome ! I always make a point of matching any code style policy or coding preferences when contributing, so no problem.

My initial patch was worked out in a few minutes... It probably needs some refining.

mobarre:
No, I have not done this for a reason, and the suggestions are very welcome ! My initial patch was worked out in a few minutes... It probably needs some refining.
Would this do: a single test in gps_driver.h + a define and ifdefs in the cpp file ? or even better a test through cmake and only ifdefs in gps_driver.cpp ?

indi-gpsd/gps_driver.cpp
```diff
@@ -279,8 +282,11 @@ IPState GPSD::updateGPS()
             break;
     }
 
-
+#if GPSD_API_MAJOR_VERSION >= 10
+    if (gpsFix->status == STATUS_NO_FIX)
```
jpaana, line 286: This change is bogus as the pointer is not initialized anywhere. The status field has indeed moved so the check should be:
if (gpsData->fix.status == STATUS_NO_FIX)

PR #190, opened 2020-08-14
Optionally clear CMOS before exposure

const86:
CMOS is normally reset just before every exposure. But sometimes residual charge may remain
and pollute new image. Running dummy short exposures just before starting a real one is found
to help in some situations. For example, it reduces ghosts of stars after dithering.
Disabled by default, user may choose to blink any number of times.

Validated on my ASI294MC Pro. Each blink takes about 300-400ms of extra time.

See extensive discussion here: https://www.cloudynights.com/topic/662887-not-rbi-on-cmos-but-this-looks-similar/

### DIFF
--- a/indi-asi/asi_ccd.cpp
+++ b/indi-asi/asi_ccd.cpp
@@ -263,7 +263,8 @@ bool ASICCD::initProperties()
                        ISR_1OFMANY, 60, IPS_IDLE);
 
     IUFillNumber(&BlinkN[0], "BLINK_TIMES", "Blinks before exposure", "%2.0f", 0, 100, 1, 0);
-    IUFillNumberVector(&BlinkNP, BlinkN, 1, getDeviceName(), "BLINK", "Blink", CONTROL_TAB, IP_RW, 60, IPS_IDLE);
+    IUFillNumber(&BlinkN[1], "BLINK_DURATION", "Blink duration", "%2.3f", 0, 60, 0.001, 0);
+    IUFillNumberVector(&BlinkNP, BlinkN, NARRAY(BlinkN), getDeviceName(), "BLINK", "Blink", CONTROL_TAB, IP_RW, 60, IPS_IDLE);
 
     IUSaveText(&BayerT[2], getBayerString());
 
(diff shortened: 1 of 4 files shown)


indi-asi/asi_ccd.cpp
```diff
@@ -1023,6 +1036,51 @@ bool ASICCD::StartExposure(float duration)
 {
     ASI_ERROR_CODE errCode = ASI_SUCCESS;
 
+    long blinks = BlinkN[0].value;
+    if (blinks > 0)
+    {
+        LOGF_INFO("Blinking %ld time(s) before exposure", blinks);
+
+        const long duration = 0;
+        errCode = ASISetControlValue(m_camInfo->CameraID, ASI_EXPOSURE, duration, ASI_FALSE);
```
knro, line 1045: This is using duration which is the exposure time? isn't this suppose to be short 100ms or something?

const86: Yes, it is exposure time. 0 seems to work for me. I have zero experience with ASI and CMOS idiosyncrasies... so I can't really prove it is the best one. Probably, I can make it configurable too.

const86: ...done.

const86:
> Thank you! Need to add BlinkNP to saveConfig function as well so that it is remembered.

Done.

indi-asi/asi_ccd.cpp
```diff
@@ -2219,5 +2220,8 @@ bool ASICCD::saveConfigItems(FILE *fp)
     if (VideoFormatSP.nsp > 0)
         IUSaveConfigSwitch(fp, &VideoFormatSP);
 
+    if (BlinkNP.nnp > 0)
```
knro, line 2223: That's always going to be 1. You probably meant
if (BlinkN[0].value > 0)

indi-asi/asi_ccd.h
```diff
@@ -154,7 +154,7 @@ class ASICCD : public INDI::CCD
         uint8_t rememberVideoFormat = { 0 };
         ASI_IMG_TYPE currentVideoFormat;
 
-        INumber BlinkN[1];
+        INumber BlinkN[2];
```
knro, line 157: It's recommended to add enums to make this clearer.

enum
{
  BLINK_COUNT,
  BLINK_DURATION
};

then in code when you use BlinkN[BLINK_COUNT].value for example, it's clear to which element we are referring to.

const86: Done

const86:
> This change also required a version change in CMakelists.txt and another in debian/indi-asi/changelog to mark the changes.

Done.

PR #191, opened 2020-08-14
Correct fix status check for new gpsd

jpaana:
Previous fix was totally bogus as it just dereferenced an uninitialized pointer.


knro:
So had nothing to do with version incompatibilities???

jpaana:
There is valid API version incompatibility issue as the fix status field was moved from gps_data_t to gps_fix_t (which is included in gps_data_t so the fix it trivial), but the previous fix (#188) was incorrect and just caused a segmentation fault due to just checking for a pointer that wasn't initialized anywhere.

PR #203, opened 2020-08-21
FIX: Parse correctly Fuji shutterspeed given in minutes.

muessel:
Libgphoto2 does not support bulb mode on Fuji cameras. As of versions 2.5.25 fixed shutterspeed >60s are reportet in minutes (1m, 2m, 4m etc.).

The 'm' postfix ist just ignored by indi_gphoto_ccd. This is fixed in this patch.

I've sucessfully tested it on my X-T2 with exposure times up to 15 minutes.


knro:
Thank you, that's great!

PR #205, opened 2020-08-23
Pentax shutter speed fixes

karlrees:
- Fixed problems with shutter speeds of less than a second
- Block shutter speeds of less than a second in bulb mode to prevent hanging the camera
- Defined a valid minimum shutter speed, which prevents problems with Bias frames in Ekos
- README updated and other minor fixes

### DIFF
--- a/indi-pentax/README.md
+++ b/indi-pentax/README.md
@@ -19,14 +19,15 @@ To build/install everything manually, first clone the indi-3rdparty repository:
 ```
 mkdir -p ~/Projects
 cd ~/Projects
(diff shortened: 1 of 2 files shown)


karlrees:
It might also be time to create a separate page for the native Pentax driver on the main website (under https://www.indilib.org/ccds.html).  The README would be a good starting point, and I'd be happy to supply any other additional info.

knro:
@karlrees Absolutely, you have access to add/edit entries. Maybe start with the INDI GPhoto one and customize it as necessary?

PR #222, opened 2020-09-30
Interferometer

iliaplatone:


knro:
Can you please indicate what's the content for this PR?

iliaplatone:
I added a new library: libxc (libahp_xc.so), which allows higher-level usage of the cross-correlator. I'm working on spectra and time correlations, for radial velocity analysis via radio or optical wavelengths.
libahp_xc helps in this with higher-level functions and it's usable from third-party applications.
P.S.: since I updated the firmware part, it is simple for INDIlib to use an external library. I don't want to modify the firmware repository anymore, but it allows variants with different chips and subsequently different number of lines, delay sizes and spectra sizes.

iliaplatone:
Only indi-ahp-correlator and libxc folders were modified, added entries into debian folder and wrote into CMakeLists.txt

xsnrg:
This broke the Fedora packager build with the introduction of the new library. A .spec file is needed. I'll whip up a PR to add it shortly. This is only mentioned for awareness.

xsnrg:
Other than the config at Copr to package libxc, The spec file in the requiring package (indi-ahp-correlator) needs a `BuildRequires libxc` so that the environment used to build indi-ahp-correlator has the new library.  PR coming.

xsnrg:
There is a problem. libxc is already a bundled package in Fedora:

Name         : libxc
Version      : 4.3.4
Release      : 1.fc32
Architecture : x86_64
Size         : 2.3 M
Source       : libxc-4.3.4-1.fc32.src.rpm
Repository   : updates
Summary      : Library of exchange and correlation functionals for density-functional theory
URL          : http://www.tddft.org/programs/octopus/wiki/index.php/Libxc
License      : MPLv2.0
Description  : libxc is a library of exchange and correlation functionals. Its purpose is to
             : be used in codes that implement density-functional theory. For the moment, the
             : library includes most of the local density approximations (LDAs), generalized
             : density approximation (GGAs), and meta-GGAs. The library provides values for
             : the energy density and its 1st, 2nd, and (for the LDAs) 3rd derivatives.

If this is the same library with a compatible version, no problem, I can tell the Copr build to pull it in. If it is different, then there is a naming conflict that we need to figure out.  I will see what I can find. Where did you get the version you are including?

xsnrg:
@iliaplatone Looking at the header, I see it is a library of your own making.  Can we rename the package to libahp_xc instead of libxc to remove the conflict with the existing libxc: https://www.tddft.org/programs/libxc/ ?

iliaplatone:
Yes, rename it into libahp_xc. I meant to add more libraries for other products, so the libahp_ prefix is the best way to do this

xsnrg:
https://github.com/indilib/indi-3rdparty/pull/223

PR #234, opened 2020-11-13
SV305PRO detection fixed

thx8411:
Hello,

The SV305 PRO detection is fixed.
The camera is now able to guide using the ST4 port.

Best regards,

Blaise


tlkw79:
SV305 pro detection is done but still not usable. :( Even it still cannot connect the SV305 PRO under the secondary camera option in PHD2.

PR #254, opened 2020-12-09
Adding weewx support

ennui2342:
I have added basic integration with weewx weather station to treat weather radio as a weewx weather station. Data is read from serial by a weewx driver. Also included is a skin which creates static json files every ARCHIVE cycle which can be used by the indi library in place of the direct arduino access by creating a static website - example nginx config included (this isn't ideal for realtime, see readme for discussion). Given the potential for considerable other weather data and calculations to be aggregated by weewx, the ideal here would be to also include a indi_weewx driver, but my C# isn't up to working on that code :(


knro:
@sterne-jaeger Can you please check?

knro:
Right, it would be ideal to add a separate driver for this. It's in C++, not C#, but I guess that's still an issue for you?

ennui2342:
> Right, it would be ideal to add a separate driver for this. It's in C++, not C#, but I guess that's still an issue for you?

Yeah sorry Jasem, I'd love to contribute more to the great work you guys do, but my competence maxes out at mucking around with Arduino level C I'm afraid.

sterne-jaeger:
Sure, happy to see weather radio growing.

PR #263, opened 2020-12-16
Clear Bayer flag when MonoBin is active

const86:


knro:
Thanks, I don't think we need to query these values every time, we already have the latest values in the properties themselves. The code would be simplified and no unnecessary SDK calls are made.

const86:
> Thanks, I don't think we need to query these values every time, we already have the latest values in the properties themselves. The code would be simplified and no unnecessary SDK calls are made.

Well, OK. I thought about it and chose the other way randomly :)
How are INDI and ASI states synchronized? I mean, if I start exposure with MonoBin enabled and then switch if off in the middle of exposure, I should still regard it as enabled somehow.

knro:
Right, you can create a new member variable to keep track of monobin and only update it when capture is not in progress. The goal is reduce any unnecessary overhead by the extra calls to the SDK. If exposure is set to 0.001s then each call matters.

knro:
I'll merge this for now and check the performance issues, it might not be a big issue after all.

PR #265, opened 2020-12-17
asi_ccd/refactoring - using std containers

pawel-soja:
Hi!

I am getting to know the library and I made a few simplifications.

- Using std::unordered_map
- Using foreach instead of for
- Replacing a function in a simulation
- Changing the names of cameras in the simulation.
- Using the std::swap function

Regards
Pawel


knro:
Thank you, it's great to see modern C++ being utilized here. Where is the foreach part though?

pawel-soja:
> Where is the foreach part though?

What do you mean?
I got rid of the cleanup function completely. The memory is freed by the containers.
And in some places I used `for ( : )`

PR #270, opened 2020-12-23
Remove CCD_HAS_BAYER property for Mono camera

hparksy:
Fixed a bug in which Mono images were processed as color images because the CCD_HAS_BAYER property was set in QHY Mono Camera.

### DIFF
--- a/indi-qhy/qhy_ccd.cpp
+++ b/indi-qhy/qhy_ccd.cpp
@@ -1040,14 +1040,21 @@ bool QHYCCD::Connect()
         //if(ret != QHYCCD_ERROR && ret != QHYCCD_ERROR_NOTSUPPORT)
         if (ret != QHYCCD_ERROR)
         {
+            cap |= CCD_HAS_BAYER;
+
             if (ret == BAYER_GB)
                 IUSaveText(&BayerT[2], "GBRG");
             else if (ret == BAYER_GR)
                 IUSaveText(&BayerT[2], "GRBG");
             else if (ret == BAYER_BG)
                 IUSaveText(&BayerT[2], "BGGR");
-            else
+            else if (ret == BAYER_RG)
                 IUSaveText(&BayerT[2], "RGGB");
+            else
+            {
+                IUSaveText(&BayerT[2], "");
+                cap &= ~CCD_HAS_BAYER;
+    	    }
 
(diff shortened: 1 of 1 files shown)


knro:
In #1039, the call is made for CAM_COLOR, so if it is true, then it must be color and not MONO.

hparksy:
Yes your point is correct.

But unfortunately, even if it is MONO Camera, that routine will be
performed.

I tested it on my QHY600 Mono Camera.

Before modifying that routine, the QHY600 Mono's image is treated as Color.
After modifying that routine, it proceeds normally.

This may be due to a bug in
"ret = IsQHYCCDControlAvailable(m_CameraHandle, CAM_COLOR);"


2020년 12월 24일 (목) 오전 1:53, Jasem Mutlaq <notifications@github.com>님이 작성:

knro:
Ok fair enough, let's do this until QHY fixes the SDK.

hparksy:
Thank you.

I will give feedback if there are any new changes.

2020년 12월 24일 (목) 오전 2:10, Jasem Mutlaq <notifications@github.com>님이 작성:

knro:
is this error with latest QHYSDK from INDI? I just doubled checked the SDK and it does not appear to return QHYCCD_OK for this function.

knro:
Nevermind, I found the issue and will report it now to QHY for confirmation.

hparksy:
I have now tested it on Ubuntu (X86_64) and Ubuntu RaspberryPI4 (AARCH64)
with the latest SDK.

On X86_64, it returns correct results.
However, Ubuntu RaspberryPI4 (AARCH64) returns incorrect results.

The camera is QHY600.

Send the screen captured from RaspberryPI4.

I hope this helps.

2020년 12월 24일 (목) 오전 2:20, Jasem Mutlaq <notifications@github.com>님이 작성:

hparksy:
I also send the screen where the error occurred.



2020년 12월 24일 (목) 오전 3:39, Suyoung Park <hparksy@gmail.com>님이 작성:

knro:
I just pushed 20.21.23 QHYSDK release with a fix. Please re-install after git pull and re-compile indi-qhy against this SDK and report back.

hparksy:
I tested it on RaspberryPI4 (AARCH64).

The result is successful.

Thank you very much.

hparksy:
It was fixed with a fix in the SDK.

PR #273, opened 2020-12-30
indi_rpicam 1.2

lboclboc:
- Big improvement of downloading speed.
- Bug in exposure time, only whole seconds was used earlier.
- Added unit test for the mmalcamera object.


knro:
Thank you! what's vcos used for?

lboclboc:
Its part of the Video Core API:s on Raspberry so its kind of vital for indi_rpicam

QRainman:
The second camera_control.reset(new CameraControl()); in MMALDriver::Connect() crashes for non HQ cameras because mmal_connection_create() does not yield a valid connection object the second time around.

lboclboc:
Hm, ok. The thing is without it I can not get the camera to do long exposures. One manual workaround with the old version is to connect, disconnect and then connect again, take one long exposure (which will not be ok) and then take a second one, that will be ok. Something is really fishy with the vcos/broadcom code here I think ... or mine code..

I did submit a PR for a revert earlier today.

QRainman:
Unfortunately i have only a v1 and v2 camera to play around.

Also, i think the BayerConverters are suffering from a similar bug that the exposure time setting had:
cur_row[x] = static_cast<uint16_t>(byte << 2)  ==> cur_row[x] = static_cast<uint16_t>(byte) << 2
The way it is done now, the two msbs will probably be truncated.

lboclboc:
> Unfortunately i have only a v1 and v2 camera to play around.
> 
> Also, i think the BayerConverters are suffering from a similar bug that the exposure time setting had:
> cur_row[x] = static_cast<uint16_t>(byte << 2) ==> cur_row[x] = static_cast<uint16_t>(byte) << 2
> The way it is done now, the two msbs will probably be truncated.

Ok, not sure I follow. Can you elaborate a bit? I might as well fix that bug also now in that case.

QRainman:
If byte was, for example, 0xff (b1111.1111) then (byte <<2) will move everything to the left by two. Since it's an 8 bit type though there is not enough space to store the the top two bits and they are discarded, yielding 0xfc (b1111.1100) instead of 0x3fc (b11.1111.1100). Casting to 16 bits first ensures that there is enough space to actually do the << 2 op correctly, even if the number is larger than 0x3F.

PR #281, opened 2021-01-04
Added extra paths for finding header files.

lboclboc:
If libindi is compiled and installed to a non-standard path, just setting INDI_ROOT
is not enough and cmake will not find correct files from that installation without
this extra path.


knro:
non standard path like what? /opt ?

lboclboc:
Like /opt yes, or if you just want to build a debug version like in $HOME/.local/indi which was what I needed

lboclboc:
thanks

PR #288, opened 2021-01-10
Weatherradio

ennui2342:
Adding support for resistor based dew heater to weather radio.

Also a couple of minor bug fixes for DHT and the jsondocument.

I'm not really happy with the line:

`updateDewheater(dhtData.temperature, dhtData.humidity);`

since users will have to change the main file if they wanted so use say the BME sensor. But without making all the sensors proper objects I don't really know a good way to pass a reference to a general class of a temp/hum sensor in the config. @sterne-jaeger I'm happy to take some guidance here if you want this done differently.


indi-duino/devices/Firmwares/weatherradio/weatherradio.ino
```diff
@@ -483,6 +493,11 @@ void loop() {
   updateAnemometer();
 #endif //USE_DAVIS_SENSOR
 
+#ifdef USE_DEW_HEATER
+  // Change this if you want to use a different sensor for the dew heater
+  updateDewheater(dhtData.temperature, dhtData.humidity);
```
sterne-jaeger, line 498: What about shifting the dependency to the dew heater code like providing a function `updateDewheater()` in the dew heater that is called here and inside this function the call can be called with 
```
void updateDewheater() {
#ifdef USE_DHT
updateDewheater(dhtData.temperature, dhtData.humidity);
#endif // USE_DHT
...
}
```
That's not nice, but this way we keep weatherradio.ino untouched. It is quite lengthy meanwhile and I would not like to force a user to touch it

sterne-jaeger (changes requested):

sterne-jaeger:
> But without making all the sensors proper objects I don't really know a good way to pass a reference to a general class of a temp/hum sensor in the config.
I have to admit that I am not really familiar with Arduino code. I would really appreciate if all drivers are in fact objects, but as far as I understand, I need to create a library if I want to add .cpp files, right?

indi-duino/devices/Firmwares/weatherradio/weatherradio.ino
ennui2342: Yep, thats a good solution. I'll find some time to make that change and update the pull. Thanks Wolfgang.

ennui2342:
> I have to admit that I am not really familiar with Arduino code. I would really appreciate if all drivers are in fact objects, but as far as I understand, I need to create a library if I want to add .cpp files, right?

Yeah that's right. I haven't done Arduino OO myself, but I think if the library was going to be overhauled to that extent then I'd probably look at using something like https://learn.adafruit.com/using-the-adafruit-unified-sensor-driver

This would probably only be worth doing If there was an expectation to add wider sensor support in the future as it would be a major overhaul. Not sure if you have plans for weatherradio going forward or whether it's pretty much done for your needs.

sterne-jaeger (approved):
Looks good, let’s go for it!

Issue #292, opened 2021-01-13
libatik needs a min glibc version requirement


paolostivanin:
It seems that with PR #123 , the min glibc version for `libflycapture.bin` has been increased to 2.27

```
strings libatik/x64/libflycapture.bin | grep GLIBC_2.27
GLIBC_2.27
```

I think we should update the CMakeLists.txt accordingly, what do you think?

I can open a PR myself but, as I don't own an ATIK camera, I can't verify whether this new glibc min requirement is mandatory or not.

knro:
@TallFurryMan Can you please check?

TallFurryMan:
Good point, I will check. And probably also update de SDK in the process if there is a new version, it's been several months.

TallFurryMan:
The SDK is updated on my github branch, I will run a few tests and push a MR. I am not under the impression that the glibc dependency is mandatory. It seems to be an internal static link.

paolostivanin:
In case it might help, testing it requires an old Debian (stretch) or an old Ubuntu (xenial) or openSUSE 15.2 :+1:

TallFurryMan:
Tested OK on Ubuntu 18.04 32-bit.

TallFurryMan:
I sent a mail to Atik support about the C99 `powf` symbol in `libflybycapture`.

I also mentioned that, in my observatory, when the device, initially plugged in the usb port but not powered, is powered, the first driver connection will only get zero frames. The second connection will resolve the problem.

TallFurryMan:
Atik did acknowledge my question, awaiting feedback.

knro:
Any update on this issue?

knro:
@TallFurryMan Checking if there is any update on this issue from Atik?

grantbrown:
I believe [this issue](https://github.com/indilib/indi-3rdparty/issues/382) is likely related. I can use an Atik EFW 2.2 on an Ubuntu 18.04 system, but on 20.04 and newer it fails. Any ideas about a workaround?

paolostivanin:
I don't think this issue is related to your bug. Ubuntu 20.04 is using glibc 2.31 (while 18.04 was using 2.27), so this shouldn't be an issue. _Usually_, glibc is backwards compatible.

grantbrown:
[...]

PR #302, opened 2021-01-22
Update Atik SDK to 2020 October 19 release before test.

TallFurryMan:
PR integrating the October 2020 SDK from Atik.

Note that it also reverts the length parameter of `ArtemisCameraSpecificOptionGetData` to a pointer.

### DIFF
--- a/indi-atik/CMakeLists.txt
+++ b/indi-atik/CMakeLists.txt
@@ -14,7 +14,7 @@ find_package(ATIK REQUIRED)
 FIND_LIBRARY(M_LIB m)
 
 set(ATIK_VERSION_MAJOR 2)
-set(ATIK_VERSION_MINOR 7)
+set(ATIK_VERSION_MINOR 8)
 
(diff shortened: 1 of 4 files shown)


TallFurryMan:
This is in the context of issue #292. This PR is not tested on my side yet.

The only dependency on `GLIBC_2.27` comes from symbol `powf` in `libflybycapture.bin`:
```
 readelf -s ./x64/libflycapture.bin | grep 'GLIBC_2.27'
   298: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND powf@GLIBC_2.27 (26)
```

Yes, this C99 symbol was introduced in the previous release integration. Would it be a good idea that I asked Atik to revert to the older `pow` with a cast? If we go on, what would be the right way to declare that dependency in this PR.

TallFurryMan:
It appears I didn't add the MacOS dynlib :( I will push a hotfix.

PR #304, opened 2021-01-24
Add Volts event recognition for V3.4.0+ firmware

ken-self:
In response to PK's issue https://indilib.org/forum/general/8691-going-to-move-over-to-the-dark-side-for-a-while.html 
As far as I can tell the only change introduced in 3.4.0 of nexdome that may impact operation is the :Volts event so I added that in and also added a placeholder for the B command (Battery threshold)
This may be needed to set the threshold, which affects the :Volts event being emitted
Built, installed and ran the driver with no issues
I don't have a nexdome so cannot test. Is there some way PK can test before deploying? 
Release notes for 4.0.0: https://github.com/nexdome/Firmware/releases
v4.0.0 protocol: https://github.com/nexdome/ASCOM/wiki/Firmware-Protocol
Firmware compatibilty: https://github.com/nexdome/ASCOM/wiki/Firmware-Protocol#firmware-compatibility-notes
v3.0.0 protocol: https://github.com/nexdome/ASCOM/blob/release/3.0.0/Firmware/ReadMe.pdf


knro:
Thank you. However, I don't think this would solve the problem described by PK, but let's see.

ken-self:
Thanks jasem
The protocol itself is quite simple and I lined up both the old and new version and these were the only differences as far as the documentation goes. I thought the problem might be processEvent() returning false but it never gets tested anyway. 
I'll go over the firmware code in case the documentation is wrong.

PR #306, opened 2021-01-25
Advertise BGGR pattern on DSI Color III (tentative).

TallFurryMan:
This PR is a tentative fix for the unrecognised Bayer pattern of the Meade DSI Color III.


TallFurryMan:
The commit message is wrong, this is a RGGB pattern, not a BGGR pattern.

PR #307, opened 2021-01-26
Advertise RGGB pattern on DSI Color III (tentative).

TallFurryMan:
This PR is a tentative fix for the unrecognised Bayer pattern of the Meade DSI Color III.


knro:
Thanks! Any idea how does this affect the other cameras?

@bgilsrud Any thoughts?

TallFurryMan:
There should be zero impact to other cameras, based on their reported sensor type. This change is expected to put RGGB in the FITS header as Bayer pattern, and only this.

PR #312, opened 2021-01-31
Fix incorrect property assignment in indi-dsi, causing crash.

TallFurryMan:
Fixes regression introduced by 7c4c723, which was using `IUSaveText` with a null argument.

Also properly updates the driver version.


TallFurryMan:
@knro This one is urgent to restore Meade sensors functionality :disappointed:

PR #314, opened 2021-02-04
Relocate DSI Color III Bayer pattern setup

TallFurryMan:
This PR relocates the Bayer pattern setup for the DSI Color III so that it is written to the FITS header.


TallFurryMan:
9 minutes too late :) I'll push a new PR with the version change.

PR #319, opened 2021-02-06
Read mode and overscan area support for qhy cameras

nrackwit:
This change adds previously missing camera re-initialization after read mode changes and new support for capturing overscan areas of the imaging sensor

### DIFF
--- a/indi-qhy/qhy_ccd.cpp
+++ b/indi-qhy/qhy_ccd.cpp
@@ -329,8 +329,8 @@ bool QHYCCD::initProperties()
 
 
     //NEW CODE - Add support for overscan/calibration area 
-    IUFillSwitch(&OverscanAreaS[0], "INDI_ENABLED", "Include", (IgnoreOverscanArea ? ISS_OFF : ISS_ON));
-    IUFillSwitch(&OverscanAreaS[1], "INDI_DISABLED", "Ignore", (IgnoreOverscanArea ? ISS_ON : ISS_OFF));
+    IUFillSwitch(&OverscanAreaS[0], "INDI_ENABLED", "Include", ISS_OFF);
+    IUFillSwitch(&OverscanAreaS[1], "INDI_DISABLED", "Ignore", ISS_ON);
     IUFillSwitchVector(&OverscanAreaSP, OverscanAreaS, 2, getDeviceName(), "OVERSCAN_MODE", "Overscan Area", MAIN_CONTROL_TAB, IP_RW,
                        ISR_1OFMANY, 0, (IgnoreOverscanArea ? IPS_IDLE : IPS_OK));
 
@@ -605,8 +605,7 @@ bool QHYCCD::updateProperties()
                     ReadModeN[0].value = currentReadMode;
                     
                     //NEW CODE - Query and display read mode name for more informative console output 
-                    char currentReadModeName[255];
-                    strcpy(currentReadModeName,"");
+                    char currentReadModeName[255] = {0};
                     int retVal = GetQHYCCDReadModeName(m_CameraHandle, currentReadMode, currentReadModeName);
                     if (retVal == QHYCCD_SUCCESS) {
                         LOGF_INFO("Current read mode is: %s", currentReadModeName);
@@ -1852,7 +1851,11 @@ bool QHYCCD::ISNewSwitch(const char *dev, const char *name, ISState *states, cha
             bool isIgnored = (OverscanAreaS[INDI_ENABLED].s == ISS_OFF); //Overscan Area included switch is 'off', thus excluded
             
             if (isIgnored == IgnoreOverscanArea)
+            {
+                OverscanAreaSP.s = IPS_OK;
+                IDSetSwitch(&OverscanAreaSP, nullptr);
                 return true;
+            }
 
             IgnoreOverscanArea = isIgnored;
             
@@ -1860,8 +1863,6 @@ bool QHYCCD::ISNewSwitch(const char *dev, const char *name, ISState *states, cha
             OverscanAreaS[INDI_ENABLED].s = IgnoreOverscanArea ? ISS_OFF : ISS_ON;
 
             OverscanAreaSP.s = IgnoreOverscanArea ? IPS_IDLE : IPS_OK;
-
-            IDSetSwitch(&OverscanAreaSP, nullptr);
 
             if (IgnoreOverscanArea)
             {
@@ -2067,7 +2068,7 @@ bool QHYCCD::ISNewNumber(const char *dev, const char *name, double values[], cha
                 
                 currentReadMode = newReadMode;
                 
-                char currentReadModeName[255];
+                char currentReadModeName[255] = {0};
                 strcpy(currentReadModeName,"");
                 int retVal = GetQHYCCDReadModeName(m_CameraHandle, currentReadMode, currentReadModeName);
                 if (retVal == QHYCCD_SUCCESS) {
--- a/indi-qhy/qhy_ccd.cpp
+++ b/indi-qhy/qhy_ccd.cpp
@@ -327,6 +327,13 @@ bool QHYCCD::initProperties()
     IUFillSwitchVector(&CoolerModeSP, CoolerModeS, 2, getDeviceName(), "CCD_COOLER_MODE", "Cooler Mode", MAIN_CONTROL_TAB,
                        IP_RO, ISR_1OFMANY, 0, IPS_IDLE);
 
+
+    //NEW CODE - Add support for overscan/calibration area 
+    IUFillSwitch(&OverscanAreaS[0], "INDI_ENABLED", "Include", (IgnoreOverscanArea ? ISS_OFF : ISS_ON));
+    IUFillSwitch(&OverscanAreaS[1], "INDI_DISABLED", "Ignore", (IgnoreOverscanArea ? ISS_ON : ISS_OFF));
+    IUFillSwitchVector(&OverscanAreaSP, OverscanAreaS, 2, getDeviceName(), "OVERSCAN_MODE", "Overscan Area", MAIN_CONTROL_TAB, IP_RW,
+                       ISR_1OFMANY, 0, (IgnoreOverscanArea ? IPS_IDLE : IPS_OK));
+
     /////////////////////////////////////////////////////////////////////////////
     /// Properties: Utility Controls
     /////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


indi-qhy/qhy_ccd.cpp
```diff
@@ -1937,45 +2043,50 @@ bool QHYCCD::ISNewNumber(const char *dev, const char *name, double values[], cha
...
+            rc = SetQHYCCDReadMode(m_CameraHandle, newReadMode); // [NEW CODE] declaration int rc removed
             if (rc == QHYCCD_SUCCESS)
             {
-                LOGF_INFO("Read mode updated to %.f", ReadModeN[0].value);
-                // Get resolution
-                GetQHYCCDReadModeResolution(m_CameraHandle, ReadModeN[0].value, &imageRMw, &imageRMh);
+                
+                currentReadMode = newReadMode;
+                
+                char currentReadModeName[255];
```
knro, line 2070: char currentReadModeName[255] = {0};

indi-qhy/qhy_ccd.cpp
```diff
@@ -1937,45 +2043,50 @@ bool QHYCCD::ISNewNumber(const char *dev, const char *name, double values[], cha
...
+                strcpy(currentReadModeName,"");
```
knro, line 2071: Not needed, see commend above.

indi-qhy/qhy_ccd.cpp
```diff
@@ -327,6 +327,13 @@ bool QHYCCD::initProperties()
     IUFillSwitchVector(&CoolerModeSP, CoolerModeS, 2, getDeviceName(), "CCD_COOLER_MODE", "Cooler Mode", MAIN_CONTROL_TAB,
                        IP_RO, ISR_1OFMANY, 0, IPS_IDLE);
 
+
+    //NEW CODE - Add support for overscan/calibration area 
+    IUFillSwitch(&OverscanAreaS[0], "INDI_ENABLED", "Include", (IgnoreOverscanArea ? ISS_OFF : ISS_ON));
+    IUFillSwitch(&OverscanAreaS[1], "INDI_DISABLED", "Ignore", (IgnoreOverscanArea ? ISS_ON : ISS_OFF));
```
knro, line 333: IgnoreOverscanArea is always true by the time we reach this code. So just set the first switch to ISS_OFF and second switch to ISS_ON

indi-qhy/qhy_ccd.cpp
```diff
@@ -1778,6 +1840,50 @@ bool QHYCCD::ISNewSwitch(const char *dev, const char *name, ISState *states, cha
...
+        //////////////////////////////////////////////////////////////////////
+        //NEW CODE - Add support for overscan/calibration area 
+        else if (!strcmp(OverscanAreaSP.name, name))
+        {   
+            
+            IUUpdateSwitch(&OverscanAreaSP, states, names, n);
+            bool isIgnored = (OverscanAreaS[INDI_ENABLED].s == ISS_OFF); //Overscan Area included switch is 'off', thus excluded
+            
+            if (isIgnored == IgnoreOverscanArea)
+                return true;
```
knro, line 1855: Before "return true" you must update the property, otherwise, it would remain as IPS_BUSY, so

if (isIgnored == IgnoreOverscanArea)
{
   OverscanAreaSP.s = IPS_OK;
   IDSetSwitch(&OverscanAreaSP, nullptr);
   return true;
}

indi-qhy/qhy_ccd.cpp
```diff
@@ -1778,6 +1840,50 @@ bool QHYCCD::ISNewSwitch(const char *dev, const char *name, ISState *states, cha
...
+            } else { 
+                //Image Info 
+                SetCCDParams(sensorROI.subW, sensorROI.subH, PrimaryCCD.getBPP(), PrimaryCCD.getPixelSizeX(), PrimaryCCD.getPixelSizeX());
+                //Image Settings
+                UpdateCCDFrame(sensorROI.subX ,sensorROI.subY, sensorROI.subW, sensorROI.subH);
+            }
+
+            LOGF_INFO("The overscan area is %s now. The effective frame starts at coordinates (%u, %u) ", IgnoreOverscanArea ? "ignored" : "included", IgnoreOverscanArea ? effectiveROI.subX : 0, IgnoreOverscanArea ? effectiveROI.subY : 0);
+            
+            IDSetSwitch(&OverscanAreaSP, nullptr);
```
knro, line 1884: Why are you setting the OverscanAreaSP switch twice here? One should be enough the at the end.

nrackwit:
I am creating this pull request per Jasem's request. It is my first time working with git, please bare with me. I have used the code so far on my QYH294M-Pro and on a QHYIII5-290M and only for a few hours. Please advise how to best support additional testing.

knro (changes requested):
Some minor changes. Couple of questions:

1. Did you test QHY video streaming with this change?
2. What other cameras have you tested this with?

Thank you!

indi-qhy/qhy_ccd.cpp
nrackwit: Yes, much better. It should be updated in the new commit.

indi-qhy/qhy_ccd.cpp
nrackwit: Yes, it's also updated in the last commit.

indi-qhy/qhy_ccd.cpp
nrackwit: Also updated..

indi-qhy/qhy_ccd.cpp
nrackwit: Yes, good find. That was a copy-and-paste leftover,  thanks for seeing it.  It is also updated in the last commit.

nrackwit:
I haven't tested the code with streaming yet. I wonder if it makes sense at all to allow streaming mode for the non-standard read mode resolutions and I have yet to test how the camera behaves in that case. What are your thoughts?

nrackwit:
I confirmed that streaming is working with the new code on the smaller QHYIII5-290M. I wasn't able to successfully get the QYH294M to stream yet, although stream modes in both read modes are supported per camera specification.   For different resolutions, i.e. full resolution in 11M mode or 1024x768 at 2fps, the stream will opens but not reach the end condition of my 10s timed stream or 30 frame limited stream.  I have to reconnect the camera to stop and close the stream. This though also is the same behavior with Indi 1.8.8. Thus at this time I suspect it has more to do with me not having the all the camera, frame and stream settings lined up yet. I will keep looking.

knro:
I think we need to add an option to downsample the stream. This is is only way to support reasonable fast streams given all the limitations of the framework.

nrackwit:
I've been making some progress, my code now also allows streaming with the 294M. It seems the current code hasn't been correctly initializing the camera for streaming and then also during tear-down doesn't put it back out of streaming mode.  What adds to it is that the camera itself only seems to go into streaming mode for some of the  resolutions it is specified for but stops communicating for others.  

QHY294M streaming 1920x2160 @ 2fps

knro:
Great! Let me know once you have a working solution so I can test with the cameras I have here.

nrackwit:
Have a look this branch [indi-qhy-streaming-work](https://github.com/nrackwit/indi-3rdparty/tree/indi-qhy-streaming-work/indi-qhy). I have been testing against SDK v20.12.27 and 21.02.01 and with my camera it works well for any 4:3 or 16:9 frame format. I can't seem to be able to obtain any frames when streaming in the native chip resolution, though. Not sure why. That's something that is neither working with the video tester nor the live frame tester that comes with the SDK.

I haven't checked but I think the driver code only picks up the custom frame format when specified in the image settings but not when using the frame properties in the streaming tab.

knro:
Thank you, I'll update the QHYSDK so please try to test with that as well.

knro:
QHYSDK 21.2.11 is up now.

nrackwit:
OK, where are you able to get that version? On the [QHY SDK release page](https://www.qhyccd.com/html/prepub/log_en.html#!log_en.md) the newest download I can spot is still the 21.02.01.

knro:
The one in INDI 3rdparty is v21.2.11 which I built today.

nrackwit:
Ahh ok. I was able to pull and install from today's master and all still works but I'm a bit confused, the sdk version I am getting from those binaries on my Mac is still the older 20.12.11:

SDK_indi-3rdparty_master

Before I did the install I ran a distclean from the QHY sdk distribution which would have removed any previously installed sdk. That step did remove the files under /usr/local/lib/indi/DriverSupport/qhy/firmware/ and I checked with
"sudo find / -type f -name QHY*.HEX" that there aren't any other installs that Indi would find.

The new installed binaries have a Feb 9th timestamp, so that looks right ([terminal.txt](https://github.com/indilib/indi-3rdparty/files/5969169/terminal.txt)).

knro:
Thank I've just updated the Mac binaries.

nrackwit:
OK thank you, I haven't gotten to testing again but will take a look later this week. I'm planning to complete this change and submit a PR over the coming weekend.

knro:
I've received some feedback regarding ReadMode from QHY, can you please check it out?

> The readmode requires to be set before the initQHYCCD. Because some camera's readmode may change the sensor resolution. Like the QHY294PRO has both 11MP mode and 47MP mode. And the QHY42PRO has the 2k*2k standard mode and 4k*2k HDR mode (one image output two channel . two 2K*2k image combine to 4k*2k).   But I tested the QHY294MPRO 47mp in Kstars but it is still 11mp resolution. So I think the readmode should not be set in correct sequence.
> 
> For readmode and for single frame / live frame mode. If camera is working for frame already, it need to stop and close for first then set readmode and streamming mode and intiQHYCCD agin. 
> 
> ........
> CloseQHYCCD(camhandle); 
> ReadMode set
> steamming mode set
> InitQHYCCD(camhandle);
> ......

nrackwit:
Thank you Jasem for sharing QHY's feedback. I am doing the camera init but not yet doing the call to CloseQHYCCD(camhandle). Let me try and test this.

nrackwit:
Calling CloseQHYCCD() going out of streaming mode or exposure mode doesn't work too well. It would just disconnect the camera and stop Indi. It seems to work well by just calling InitQHYCCD() when going into a streaming mode change either from streaming to exposure or vice versa.

Test sequence working:
1) Connect Camera
2) Take exposure
3) Start Streaming (calls QHYInit(), then sets bit depth due to change of  streaming mode property)
4) Stop Streaming
5) Take exposure (calls QHYInit(), then sets bit depth due to change of  streaming mode property)
6) ...

Test sequence not working:
1) Connect Camera
2) Take exposure
3) Start Streaming (calls CloseQHYCCD(), then calls QHYInit(), then sets bit depth due to change of  streaming mode property) --> FAILS
4) Stop Streaming
5) Take exposure (calls CloseQHYCCD(), then calls QHYInit(), then sets bit depth due to change of  streaming mode property)
...

PR #321, opened 2021-02-06
Fix race in ASICCD::StopStreaming

nadvornik:
I got some crashes on stopping streaming with 1sec exposure.
I think this change fixes it.


knro:
Any sides effects detected with this change?

PR #322, opened 2021-02-06
Enable tcp connection in indi-duino

nadvornik:
TCP connection has been tested on ESP8266 board with this firmware:
https://github.com/firmata/arduino/tree/master/examples/StandardFirmataWiFi


knro:
Thank you. Was this tested with the Radio Weather as well?

nadvornik:
As far as I understand, Radio Weather is independent code and this change does not affect it.

PR #327, opened 2021-02-13
Allow calibrating of ZWO EFW

wgauvin:
Adding the ability to calibrate the ZWO EFW

Have tested this with my own EFW, it does run the calibration, though the calibration is an async process so the call to the `libasi` call just returns success. From the sound of the EFW the calibration takes 10-15 secs (the ZWO docs say it could take 60 secs), might have to check if the status of the EFW to see if it's moving and only after it has stopped that I could say that the calibration is done.

The motivation for this change was that I need to calibrate the EFW but didn't have a windows machine to setup the ASCOM driver to run the calibration.  I used the `gpsd` refresh functionality as a basis for what was needed to allow this functionality.

### DIFF
--- a/indi-asi/asi_wheel.cpp
+++ b/indi-asi/asi_wheel.cpp
@@ -351,25 +351,7 @@ bool ASIWHEEL::ISNewSwitch(const char *dev, const char *name, ISState *states, c
         }
         if (!strcmp(name, CalibrateSP.name))
         {
-            CalibrateS[0].s = ISS_OFF;
-            CalibrateSP.s   = IPS_BUSY;
-            IDSetSwitch(&CalibrateSP, nullptr);
-
-            LOGF_DEBUG("Calibrating EFW %d", fw_id);
-            EFW_ERROR_CODE rc = EFWCalibrate(fw_id);
-
-            if (rc == EFW_SUCCESS)
-            {
-                LOGF_DEBUG("Successfully calibrated EFW %d", fw_id);
-                CalibrateSP.s = IPS_OK;
-            }
-            else
-            {
-                LOGF_ERROR("%(): EFWCalibrate = %d", __FUNCTION__, rc);
-                CalibrateSP.s = IPS_ALERT;
-            }
-            IDSetSwitch(&CalibrateSP, nullptr);
-            return true;
+            return Calibrate();
         }
     }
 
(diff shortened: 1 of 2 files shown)


knro:
Perhaps you can add a 20 secs timer and once it is done, update the status of the calibration property from BUSY to OK?

wgauvin:
@knro  - that is what I was thinking.  I'll work on that tomorrow and hopefully get it working. If you know of an example of doing the timer that would be good (trying to get my head around the INDI framework so some pointing in the right direction will be very benefical)

knro:
You can use IEAddTimer(20000, &ASIWHEEL::setCalibrationCompleteHelper, this);

Then create static void setCalibrationCompleteHelper(void *context) which when called will invoice another function.. i.e.

static_cast<ASIWHEEL>(context)->setCalibrationComplete()

Then in this function, you'll simply set the status to IPS_OK and update the property.

wgauvin:
20 seconds didn't work as it was too quick (last test it took 34 seconds).  I did however find while it calibrates the status of commands is `EFW_ERROR_MOVING` but if you call `EFWGetPosition` while it is moving the position value is -1. So I've implemented a simple do/while loop that sleeps at the polling rate to keep checking the EFW for the state. I've tried this locally with my EFW.

I also note that after calibration that `libasi` puts the EFW back into the currently selected slot without the need for us to call `SelectFilter`.

Here is the output of my indiserver log
```
2021-02-14T02:50:00: Driver indi_asi_wheel: read setSwitchVector ASI EFW FILTER_CALIBRATION Busy
  CALIBRATE='Off'
2021-02-14T02:50:00: Driver indi_asi_wheel: read setNumberVector ASI EFW FILTER_SLOT Busy
 FILTER_SLOT_VALUE='4'
2021-02-14T02:50:00: Driver indi_asi_wheel: read message ASI EFW   '[DEBUG] Calibrating EFW 0'
2021-02-14T02:50:34: Driver indi_asi_wheel: read message ASI EFW   '[DEBUG] Successfully calibrated EFW 0'
2021-02-14T02:50:34: Driver indi_asi_wheel: read setSwitchVector ASI EFW FILTER_CALIBRATION Ok
  CALIBRATE='Off'
2021-02-14T02:50:34: Driver indi_asi_wheel: read setNumberVector ASI EFW FILTER_SLOT Ok
 FILTER_SLOT_VALUE='4'
```

knro:
Wouldn't it be better to check this in TimerHit? the while/loop would make the driver completely unresponsive until the calibration is complete. While this might be indeed desired, it might be better to just perform this check in TimerHit()

wgauvin:
@knro  - thanks for you input.  I've made a change to use a timer to poll. I've tested this locally on my server with the EFW.

PR #329, opened 2021-02-15
Driver for RPi GPIOfor power distribution

ken-self:
Specifically, configured to the GPIO pins used by ASI Air Pro devices. Allows 4 GPIO pins to switch power on/off or use PWM control. Requires pigpio to be installed and running as daemon pigpiod. pigpio provides PWM control over GPIO but needs to run as root to access GPIO pins.
Allows owners of ASI Air Pro to install alternative OS distributions e.g. (Ubuntu 64 bit, Astroberry) and still make use of the power ports on the device. Allows others to add power control using e.g. MOSFET devices controlled from GPIO 12, 13, 18, 26

If not suitable for INDI lib, I can continue to host source code from my github repo


knro:
Thank you. It's not possible to directly use [wiringpi](http://wiringpi.com) or [pigpio C interface](https://github.com/joan2937/pigpio)? That would be preferred then relying on some Python dependency.

ken-self:
Hi Jasem
There is no Python dependency except in the install of pigpio. That dependency only exists to provide the Python interface to the libraries. We could skip that part of the install.
I tried the pigpio C interface but it requires the driver to run with root access. Sames goes for wiringpi. The GPIO functionality in Astroberry overcomes this with a UDEV rule  but I think that only works because libgpiod is in kernel space. (Not an area I understand very well). Also wiringpi only seems to support PWM on one pin but we want to provide PWM on 4 pins.
So there is a dependency on the pigpiod daemon running as root but that is also written in C. I could use the direct interface if someone smarter than me can work out how to set the required permissions. I assume we dont want to run the driver as root.

knro:
Right we do not want to run as root. We also desire that folks installing this driver would not need to run any additional scripts..etc. If it can be handled completely form the debian package, then all the better.

ken-self:
Understood. This might get tricky. It seems that there is a debian package but it does not include the pgpiod daemon.  Explained in this issue: https://github.com/joan2937/pigpio/issues/247
So the debian packages include only the client side components. What is missing is the pigpiod daemon and the pigpio library (for direct GPIO access). We can ignore pigpio
While I've been researching I've created a libpigpio to build the libraries and daemon and excluded most of extraneous components. I can  reduce this further to just the pigpiod daemon as everything else can be installed with:
sudo apt install libpigpio-if-dev
sudo apt install libpigpio-if2-1
sudo apt install pigpio-tools
I also checked the licencing and the code is completely unencumbered: https://github.com/joan2937/pigpio/blob/master/UNLICENCE

So is it acceptable to install the pigpiod daemon program with the driver?  I have been told that the program is already installed with Astroberry but I want to confirm that myself. Either way the next question is about installing the pigpiod.service on systemd and following on from that, enabling the service. That should only happen if the user explicitly installs the driver. 

Happy to take whatever advice you have. I'm aware there are some thorny issues involved.

knro:
So the daemon is a must if we want non-roon GPIO control, correct? I suppose we can then include daemon here given it does not cause conflict with other packages?

ken-self:
Thats right. there isn't much to it. Basically an interface to the library functions. Given that the libraries are already debian packages I can build the daemon source along with the driver.

knro:
Great! Can you add the debian files for this as well? Many samples are in the debian directory

ken-self:
Sure. I'll give it a go. I like to learn these things.  I'll let you know when it's ready to go.
Currently testing that it all builds ok.

g60madman:
@ken-self I will do some testing with your new repo, I am setting up my ASI Air Pro with Ubuntu Server and pigpio is not included in Ubuntu Server. I will report back later today.

g60madman:
@ken-self @knro OK I just set up my ASI Air with Ubuntu server 20.10 following these instructions

https://ubuntu.com/tutorials/how-to-install-ubuntu-on-your-raspberry-pi#1-overview

After that was all set up with no desktop I installed the following
https://github.com/indilib/indi

followed by
https://github.com/ken-self/indi-3rdparty/tree/asipower/indi-asi-power

And then the remaining drivers I needed for my gear
https://www.indilib.org/telescopes/skywatcher/eqmod.html
https://www.indilib.org/ccds/zwo-optics-asi-cameras.html

Once that was completed I installed the INDI Webmananger and setup, my drivers
https://github.com/knro/indiwebmanager

<a href="https://ibb.co/KWCwqDP">Screenshot-from-2021-02-19-11-16-29</a>

After that, I installed Kstars on my Ubuntu 20.04 Desktop and connected to the INDI Webmanger. The power comes up and works as I can hear the fan in my ASI533MC Pro however the drivers will not connect to it no matter what I call them. I tried the defaults of the driver name and also used the name of USB device name in dmesg

<a href="https://ibb.co/9cybxfB">Screenshot-from-2021-02-19-11-13-15</a>

2021-02-19T11:05:10 Unable to establish remote devices:
+ ASI533MC Pro
+ ZWO EFF
+ ASI120MM Mini
+ EQMod Mount
Please ensure remote device name corresponds to actual device name.

2021-02-19T11:05:10 Unable to establish remote devices:
+ ASI533MC Pro
+ ZWO EFF
+ ASI120MM Mini
+ EQMod Mount
Please ensure remote device name corresponds to actual device name.

g60madman:
@ken-self Just an FYI on the drop downs for the name of the power ports I don't have an option for Mount. I am not sure if the problem is the fact my drivers are not working or not.

ken-self:
@g60madman You need to have pigpiod daemon running as root for the driver to work. There is a pigpiod.service file in the repo and instructions in the readme for setting it up as systemd service. Or jusr run in background as sudo.
As we speak I just made some changes I have not yet tested to simplify the install so you might have an unstable version. Inshort I'm trying to statically link the pigpio library into the  pigpiod daemon to avoid another library install.
re: Mount option. I wasn't aware of  that one. Can you provde detail? I assume it is just an on/off for the mount power?
Next major release I would like to provide options to nominate any GPIO and have custom labels

g60madman:
@ken-self yeah the driver does in fact work for the power, as my camera's fan does not start still I start the driver. Just having issues with the other drivers for ZWO CCD, EQMod Mount, ASI EAF. The drops downs I am talking about are like this in the KStars app

<a href="https://ibb.co/G2rFjZ5">Screenshot-from-2021-02-19-13-47-29</a>

And here are the ZWO Asi Air dropdowns. The ZWO ones will be more limited as they do not support as much hardware as INDI does

<a href="https://ibb.co/tzSSM1g">Screenshot-from-2021-02-19-13-54-08</a>

I also followed the instructions you had to build pigpio as it's not included with Ubuntu Server for Pi and pulled it down since you made the changes 10 hours ago so I am running the latest one you committed.
https://github.com/ken-self/indi-3rdparty/tree/asipower/indi-asi-power

ken-self:
@knro I've made the changes for debian then tried to run ./make_deb_pkgs indi-asi-power but got a Permission denied error on debian/rules. 
Also, when building the driver: sudo make install now successfully installs the systemd pigpiod.service, runs daemon-reload and enables and starts the service 
Once I can get make_deb_pkgs working I can see about adding that to the package

ken-self:
Close  to finishing now. Getting the errors below from the debian build.
What is the usual way to get the files into the debian/indi-asi-power directory? Is that part of the "make" process for the driver?
ken@avalon-pi3-sd64:~/Projects/indi-3rdparty$ ./make_deb_pkgs indi-asi-power
dh binary --with=systemd
   dh_update_autotools_config
   dh_testroot
   dh_prep
   dh_install
dh_install: warning: Cannot find (any matches for) "pigpiod" (tried in ., debian/tmp)

dh_install: warning: indi-asi-power missing files: pigpiod
dh_install: warning: Cannot find (any matches for) "indi_asi_power" (tried in ., debian/tmp)

dh_install: warning: indi-asi-power missing files: indi_asi_power
dh_install: warning: Cannot find (any matches for) "pigpiod.service" (tried in ., debian/tmp)

dh_install: warning: indi-asi-power missing files: pigpiod.service
dh_install: error: missing files, aborting
make: *** [debian/rules:10: binary] Error 255

ken-self:
@knro I think it is ready to go now. Tested the deb package and it installed the binaries and enabled and started the systemd service

g60madman:
@knro if you want to post it in the nightly ppa:mutlaqja/indinightly I can test it out to make sure it's working.

knro:
Great work! will do.

g60madman:
@knro I checked https://launchpad.net/~mutlaqja/+archive/ubuntu/indinightly and do not see the indi-asi-power. I assume it has not been built out yet for testing? If you can let me know when it's there I will do further testing.

knro:
It's here: https://code.launchpad.net/~mutlaqja/+recipe/indi-asi-power-nightly

First build failed because the repo wasn't updated yet on Launchpad. It should be updated in the next few mins.

g60madman:
@knro Thanks I will keep an eye on it and build it out and test it! I have my ASI Air Pro fully working with Ubuntu 20.04.2, and INDI just no power unless I add the following line into /boot/config.txt

#Enable 12v gpio output
gpio=18,12,13,26=op,dh,pu

g60madman:
@knro still not seeing anything posted yet. Let me know.

ken-self:
@g60madman the config.txt line sets the default conditions for the GPIO:
op=output, dh=drive high, pu=pull up.
We can't alter the config.txt with the install (too risky) but should be able to set those in the driver. So users would not need to edit their config file. 
Testing on my straight RPi I didn't need to do anything in config.txt to get LEDs to light up but the pins could well be set to drive low by default. 
I would be inclined to default to pull-down if the pins are drive-high so that power is off by default

g60madman:
@ken-self we don't need those lines as long as the driver works. Once I can download the driver from the ppa I can test it to verify. For now, I have been able to skip the driver install for indi-asi-power and just use the config file to power my gear.

knro:
It's failing here: https://launchpadlibrarian.net/524457240/buildlog_ubuntu-focal-amd64.indi-asi-power_0.6~202102221350~ubuntu20.04.1_BUILDING.txt.gz

debian file install should include this file prefix

ken-self:
The root cause is:
```
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
```
In CMakeLists.txt I have:
```
include(FindPkgConfig)
pkg_check_modules(SYSTEMD "systemd")
```
to test for systemd and its default directory for .service files.  So that make install would put the pgpiod.service file in the right place. 
So there is a build dependency on package pkg-config which I think was automatically installed with Ubuntu 20.04 for me

I could just drop that and put pigpiod.service anywhere that comes under debian/ and let developers manually install the service.
Or 
```
apt install pkg-config
```
I'll have a bit of a play with different options and get back to you.

ken-self:
New PR raised for the package build fix

PR #330, opened 2021-02-16
Celestronaux driver completion: cordwrap, autoguide, direct USB (alpha testing)

fabriziop:
I made an effort to complete the Celestron AUX driver adding/completing the following features:

1. cord wrap is completed setting the hourly angle origin as computed by the Alignment Subsystem.
2. autoguide is added, it is maily copied from Rick Bassham code in Celestron CGX driver.
3. direct connection to mount USB ports is added.

Unfortunately, my setup does not allow me to test this features. This code compiles and runs without fatal errors but these functionalities are completely untested.
This PR is for anyone wishing to help test the code.


knro:
@jochym Can you please check this PR?

knro:
Today is the first time I test this driver with Celestron Evolution mount. However, I couldn't get it to work. Is there a guide on how to use this with a mount?

Should the mount be aligned first? or just turned on? Should be it be in home position? If yes, what is that exactly?

In my case, I aligned the mount using the 2-star alignment. Then I connected in KStars and it was way off. I used aligned to sync it, which then worked, but any subsequent GOTO does not work and keep getting this:

2021-02-17T15:45:32: [INFO] Goto - tracking requested 
2021-02-17T15:45:13: [INFO] Goto - tracking requested 
2021-02-17T15:44:54: [INFO] Goto - tracking requested 
2021-02-17T15:44:34: [INFO] Goto - tracking requested 

I enabled logging, so I have the logs. But I think we must really clarify exactly how this driver is used so we update the documentation on it.

fabriziop:
I agree. I'll update the info in the Celestron AUX driver page and I'll add a section for using the alignment subsystem of the driver.

jochym:
@fabriziop @knro I am terribly sorry for not doing anything about the driver recently. I had simply no spare time. I'll try to return this weekend. 
But answering your direct questions Jasem:
1) The mount should be close to home position - pointing approximately at northern horizon 
2) You need to activate alignment subsystem for it to work any reasonable way
3) Sync with one star - then goto the next it should be close enough (depending on your leveling/coords) sync with it go to the third - this one should be in quite close if not in your fov sync and you should be done.

Any active (involving movement) interaction with the HC will make the alignment invalid.

As for the non-working GOTO. Was it moving but into a wrong position or not moving at all?

knro:
I tested today and here are my observations:

+ Mount assumes it's starting due SOUTH, not NORTH. This is wrong. I checked the code and found that in lines 1922 and 1923 it actually adds half a revolution to the azimuth here.. why? 
+ I commanded the mount to go to exactly west with altitude ~30 degrees. It start slewing there, arrived, then after 1 second, it started to slew slowly CCW (going south) for no apparent reason. I had to cut the power to abort it.
+ Even though I checked ALL the debug options, including Celestron AUX Verbose. I don't see the actual low-level messages logged. The Celestron AUX debug is not working.

Here is the log I was able to get, I'm afraid it doesn't contain a lot of information still due to the missing Celestron AUX Verbose log above. Now I didn't check if this issue is with the PR or it was an issue already prior to this PR, but it's there now.

[log_20-21-46.txt](https://github.com/indilib/indi-3rdparty/files/6004690/log_20-21-46.txt)

jochym:
It should make a short (approx 1deg) second goto for precission (the same way the sky safari does this). But if this just moves constantly (more than 5 deg) this is probably tracking with wrong idea of actual position. Do you have correct lat/lon/time set in site tab?

knro:
Of course, everything is correctly set. I'm worried now about the the fact it doesn't really log now. Not to mention 180 degrees off for startup position. All this should be accounted for.

jochym:
The startup position is a mess. The mount zero is one, the HC is another, sky safari yet another (I do not remember which is which). I have tried to make it work the same as skysafari. 
But you are right this should log and should be at least consistent. I'll try to carve some time for forking on PR this Saturday.

jochym:
I started working on this. It seems that the driver deteriorated why I was not paying attention. Sorry! 

This is from testing on 1.8.8 - freshly build on debian buster

@knro Regarding the movement after sync. It is caused by tracking *wrongly* trying to get to the previous (before sync) place on the sky. This is new, unexpected and wrong the position should be synced and assumed to be correct at the moment of syncing.
The initial position (after connection) should be the same with and without alignment active. Right now it jumps N<->S when you switch alignment on/off. We need to correct this.

There is also a problem with connection over tcp if you start the driver from the command line or ecos. It accepts network connection and communicates OK with the scope but simultaneously tries to connect over serial with a lot of errors in the log. If you start from device manager sa Celestron WiFi it works without the problem - not so if you use ekos or command line.

The GPS snooping seems not to work.

The PR does not compile with recent 1.8.8, so I need to switch to master for libindi.
I will continue and report the progress here.

knro:
@jochym Thank you! I also started refactoring this driver to make it more compliant with INDI standards. So hopefully tomorrow I can finish from that and then perhaps you can include your fixes? The refactor is huge.

jochym:
Great. From the initial testing of the PR it seems to be much better. It works from the command line and connects properly. The only possible problem may be the home position it is switched to S (which is actually better for me - I have no norther horizon) but the mount instructions and the Celestron software uses N as home position (after reset). I think we should make this  configurable.

The GPSD snooping issue is fixed. The Alignment warning is useful but slightly misleading - it shows up even if the alignment is active but you have no position. 
I think we should integrate this PR and your refactorings and merge it. It is already better then master.

I have also spotted a possible bug in tracking - when crossing the meridian the position (Dec) 
 becomes unstable and mount switches once a second between two close (one deg probably) positions. Repeated goto fixes this, but it is bad for the mount and indicates some mistake in our arithmetic (probably wrong wrapping angles somewhere). I'll try to hunt for this but keep your eyes open as well.

Thanks again @knro and @fabriziop for getting involved and keeping this alive.
I'll report results from ekos soon

jochym:
In ekos from kstars 3.4.3 i can start and use it ok in remote mode. If I start the drivers by ekos (local mode) the device dialog comes up with expected tabs and I get segfault and kstars crash.
In console the last line is :
```
org.kde.kstars.indi: Celestron WiFi :  "[INFO] Got echo of GET_VERSION from AZM "
org.kde.kstars.indi: Celestron WiFi :  "[INFO] Got GET_VERSION response from AZM: 7.17.98  "
org.kde.kstars.indi: Celestron WiFi :  "[INFO] Got echo of GET_VERSION from ALT "
org.kde.kstars.indi: Celestron WiFi :  "[INFO] Got GET_VERSION response from ALT: 7.17.98  "
org.kde.kstars.indi: < Celestron WiFi >: < Firmware Info >
org.kde.kstars.indi: Celestron WiFi :  "[INFO] Debug is enabled. "
Segmentation fault
```
It was the same with indi/driver compiled from 1.8.8 - this may be some local problem.

So: good work @fabriziop ! We need to work on this more but it is already much better.

knro:
@jochym I just pushed a major rework for the driver that fixes logging among other things. I didn't change any code relating to pointing. It slewed fine, but when parking it went to parking position at full speed (north), but as soon as reached there, it started going at half-max speed somewhere and I had to abort it. I can get the logs next time.

Can you please apply your changes against master now? Then we can check if all is OK before we can return to examine this PR. Let me test again after you merge in your fixes.

jochym:
That is a lot of changes @knro ! Thanks! Finally, a reworked logging - I was fighting with it in the past but never had enough time to teach myself the logging system in indi properly...
As for the after-parking movement. When you park the mount you need to switch tracking off. It was that way a year ago - somehow it must have been disabled. Try parking and switching off tracking - the problem should go away - this would indicate that I am right.
Regarding merging: I'll be busy till Friday. I'll try to work on this in spare moments.

knro:
@pawel-soja thanks for the update. Can you tell me which changes you made so I can try to apply them here? If you a diff, just attach it here and I should pick up the pace from here.

knro:
@fabriziop Can you please rebase to master? I will continue testing today.

fabriziop:
Rebased PR

fabriziop:
I am not able the reopen it.

knro:
It appears that it cannot be reopened.. can you just submit another PR? Thanks!

fabriziop:
I have to fix some merging mismatches also. It is not final. Low git skills ...

fabriziop:
OK

pawel-soja:
I was late? How can I help you now?
Most of the changes I've made are backward compatible.
If any of my changes are causing problems, feel free to drop them.
The next time I review, the files will be analyzed by me for sure.
Thanks!

jochym:
@pawel-soja You are not late. @fabriziop just moved the development to the new PR https://github.com/indilib/indi-3rdparty/pull/332

PR #331, opened 2021-02-23
indi-asi-power - issue with package build

ken-self:
Fixed the install file so that missing service file does not error.

Need advice on whether to leave the pkg-config dependency in cmake or unconditionally copy the service file to /lib/systemd/system


ken-self:
/lib/systemd/system is where package installs should put the services whereas /etc/systemd/system is where system admins should put the services.
There is also /usr/lib/systemd/system used by package installs on some distributions. On my system the contents of /usr/lib/systemd/system are identical to /lib/systemd/system so I suspect there is a symlink somewhere matching them up.
When I look into /etc/systemd/system I see a small number of services including indiwebmanager.service that I put there. And many of them are just symlinks to /lib/systemd/system so I'm guessing that was done for compatibility.

PkgConfig seems to be the "right" way to ask systemd where it expects to find services in CMake.
Apparently dh_systemd_enable puts the service into lib/systemd/service anyway. It just looks for a .service file under debian/ and does its thing. 
So the purpose of moving the service file in CMake is to put the service somewhere for dh_systemd_enable to find it and also to install it for developers building from source.
But it could just as easily be saved (or stored) in the debian/indi-asi-power directory which gets copied to debian/ (I need to test this)
So many choices....

knro:
Okay, this is a standard package, so a dependency is fine. Let me merge this and test the build again.

ken-self:
Failed again but getting closer:
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 

Interestingly in the KStars build it does find PkgConfig: https://launchpadlibrarian.net/516107625/buildlog_ubuntu-focal-amd64.kstars-bleeding_6%3A3.5.1+202101080717~ubuntu20.04.1_BUILDING.txt.gz
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1")

knro:
It needs to be added to the build dependencies in the control file.

PR #332, opened 2021-02-24
Celestronaux driver completion: cordwrap, autoguide, direct USB (alpha testing, was #330)

fabriziop:
Continuing from PR #330, accidentally closed.

Branch celestronaux_eqtrack rebased to knro  major refactor to fix mainly logging issues (7756943ccefbb12506a044bf33d713f72e87d667).

From #330:
" I made an effort to complete the Celestron AUX driver adding/completing the following features:

    cord wrap is completed setting the hourly angle origin as computed by the Alignment Subsystem.
    autoguide is added, it is maily copied from Rick Bassham code in Celestron CGX driver.
    direct connection to mount USB ports is added.

Unfortunately, my setup does not allow me to test this features. This code compiles and runs without fatal errors but these functionalities are completely untested.
This PR is for anyone wishing to help test the code. "


jochym:
I think this is just a first step. We need to fix the remaining issues. I'll return to the development over the weekend. But I think it is better to have at least partially fixed code in the main.
@pawel-soja  You were right. Fabrizio and Jasem got there faster.

knro:
@fabriziop @jochym Is there a reason for "N-based azimuth" code in the driver? In my evolution mount, it's **already** reporting correct values as is without any adjustments. Are there any mounts where this is the opposite case? I had to disable all "N-based azimuth" changes in the driver to get it working.

jochym:
I need to review that. I am fairly sure there *was* a reason. I think this is no longer the case.
It is a fairly old part of the code, so it is plausible that the "reasons" are no longer there. Can you share your changes (maybe create a branch for this?). I will work on the driver this weekend.
It was working fine few months ago. I guess the changes in the core exposed bugs in the code.

knro:
@jochym Thanks! I created a new branch with my changes that works quite well with Evolution mount: https://github.com/indilib/indi-3rdparty/tree/celestron_aux_work

I didn't test everything but basic functionality appears to work OK.

jochym:
I am reviewing your changes @knro . It seems that the N-based azimuth dates from the first work on the driver (4 years ago). Let me get it clear: what is the convention of azimuth in INDI? Did it change during past 4 years? 
If the azimuth in indi is N-based then I do not get why I put this in the code - since even comments state that Celestron uses N as zero.
The changes seem quite fine. I will compile this branch and check it on my mount. I will report the results later. Unfortunately weather does not cooperate.

knro:
Thank you for the feedback @jochym. Yesterday I tested Celestron Evolution 6" with Toupcam camera. The FOV is very small at 20'x18', but I noticed I can barely take 5 seconds image without star trails forming. Is the tracking **this bad** on an Alt-Az mount, or is this a problem with the driver itself? I usually image using equatorial mounts and you can get up to 30 secs comfortably before seeing star trails.

The definition for N-based zero azimuth never changed in INDI since 2003. You can perhaps test indoor to see if everything *behaves* as expected at least.

jochym:
That is good to know. I am only worried that I must have some reason to put these lines in.
I am testing it right now and it seems fine. Let me finish my testing and but I think it is oK to merge that variant. It would be good to have @fabriziop  test it as well. His setup is a bit different. 

Regarding tracking - with my mount (Evo 8") and  Canon DSLR (EOS 40D in main focus) it was not bad at all. I was able to consistently do 30s exposures without trails. It was quite usable.
The field rotation was a problem in high altitudes. It is definitely possible to get decent exposures out of this mount.

jochym:
I have a very windy and cloudy night. The moon is barely visible sometimes. Between the dense clouds I managed to sync on moon M44 an epsilon Boo. After that  I ran 30s exposure of M44 (SWW, 45deg Alt) high wind, f=1370mm APS-C sensor (56'x37'), quick manual focusing on the Moon. Considering the circumstances it is not bad in my opinion. See it here: https://drive.google.com/drive/folders/1_Jii4_FkuAXF5Ur8i_SDzIbh9UU4hrHi?usp=sharing

PR #335, opened 2021-02-27
LOGF_ERROR macro is expecting a char* but receiving an std::string

synoga:
The LOGF_ERROR macro is using the `INDI::Logger::print()` function through the DEBUGF macro. `print()` expects a `char *` as message (see [indilogger.h](https://github.com/indilib/indi/blob/91bcd531c4631805d24d38343692fb8acf8fff6a/libs/indibase/indilogger.h#L250))

Right now this leads me to : `pentax_ccd.cpp:532: error: cannot pass object of non-trivial type 'const std::string' (aka 'const basic_string<char>') through variadic method; call will abort at runtime`

Using c_str() fixes this.


knro:
Good catch!

PR #336, opened 2021-02-28
Fix N-based azimuth and cord wrap  handling

jochym:


jochym:
@fabriziop make sure it works on your setup please. Especially the cord wrap functionality.
@pawel-soja now is your chance ;)
@knro the cord wrap *reading* was completely broken (for a long time I'm afraid - nothing to do with your changes).

OK. @knro - merged in mid my sentence ;) Read my response regarding tracking.

ggtux:
Vielen Dank.

fabriziop:
@jochym Hello Pawel, happy to hear from you again.
This last version of the driver seems to work as usual (good).
About the cordwrap
1. My Celestron CPC seems to ignore it. I tried N and S, but nothing happens with either GOTO, manual slew or tracking enabled, crossing the limit. Perhaps, I am missing something.
2. Before your fix, there was this positive feedback https://www.indilib.org/forum/development/8246-indi-celestron-aux.html?start=120#67479
3. About your cordwrap fix: I see that you removed the offset added by function "getNorthAz". The rationale of this offset comes from remote driver operation and from the fact that my CPC scope has no remote home positioning: when the mount is powered up the telescope points to an arbitrary direction, so the encoders are zeroed there and not at leveled North for AltAz mount or at Dec = 0 and RA -> North meridian for Eq mounts.
If the driver is unaligned (zero ref. points) getNorthAz return 0, the best it can do (not ok for remote operation).
If the driver is aligned (at least 1 ref. point), getNorthAz return the offset that added to Az encoder value give zero when the scope is pointing to North. Since my CPC ignore cordwrap, it is untested.
For my setup (remote operation), the most important function of the alignment subsystem is not to correct a, hopefully, small alignment error with respect to NCP, but it is to compensate the reset of the telescope encoders at power up with an arbitrary pointing direction.

If you have any question to clarify my commits in #332 feel free to ask.

jochym:
@fabriziop (I understand your setup is equatorial):
Ad. 1) Check again with enabled serial/AUX logging - if the commands really go through and return proper values. Try with HC. If it works with HC it should work with CAUX (or we need to make it work).
Ad. 2 and 3) I am sorry I broke this use case. I did not realize it might be useful to return to true physical positions instead of initial encoder values (they indeed return to zero if you cycle power). OTOH I would actually prefer for my mount to return to the position it started in regardless of how misaligned the driver is. I think the only solution is to make it switchable.
These choices strongly depend on setup and use case and both are valid.
If you have time to implement the switch - do it. Otherwise, I will try to do it next weekend.

jochym:
@fabriziop one more note: the cord wrap *reading* code was completely non-functional in the previous versions (even after your changes). It should work now. Please check how it behaves with CPC. 
I removed the az shift bacause it seemed to interfer with proper handling of cord wrap on my setup - but it may be the artefact of errors in reading procedure I later removed. I will recheck today evening and maybe we can re-enable the ofset even without the UI.
The problem with positioning is not limited to cord wrap - it also concerns parking.

PR #341, opened 2021-03-02
Indi qhy streaming work

nrackwit:
This PR includes enhancements for QHYCCD read mode ans streaming mode support.

qyh_ccd.cpp
* Query the camera for available read modes only once during initial QHYCCD::connect() and maintain read mode information in new struct QHYReadModeInfo.
* Maintain the current read mode state in memory and call InitQHYCCD() only after the requested read mode changes from the currently set read mode. Before the camera was re-initialized also after the user set re-applied the read mode setting the camera currently is in.
*  Maintain the current streaming mode state in memory and Call InitQHYCCD() whenever camera goes into a new streaming mode change.
* Style fixes as per INDI AStyle configuration

qhy_ccd.h
* Add new variables to support the internal tracking of read mode and streaming mode states
* Add new struct and variable to hold read mode information for the various read modes of the camera
* Style fixes as per INDI AStyle configuration

qhy_ccd_test.cpp
* Add code to print out read mode information for all of the camera's supported read modes

Remaining files
* Style fixes as per INDI AStyle configuration

The change was tested with
* QHY5III-290M
* QHY290M-Pro

### DIFF
--- a/indi-qhy/qhy_ccd.cpp
+++ b/indi-qhy/qhy_ccd.cpp
@@ -860,18 +860,24 @@ bool QHYCCD::Connect()
         {
             readModeInfo[rm].id = rm;
             ret = GetQHYCCDReadModeName(m_CameraHandle, readModeInfo[rm].id, &readModeInfo[rm].label[0]);
-            if (ret == QHYCCD_SUCCESS) {
+            if (ret == QHYCCD_SUCCESS)
+            {
                 LOGF_INFO("Mode %d: %s\n", readModeInfo[rm].id, readModeInfo[rm].label);
-            } else {
+            }
+            else
+            {
                 LOGF_INFO("Failed to obtain read mode name for modeNumber: %d\n", readModeInfo[rm].id);
                 strcpy(readModeInfo[rm].label, "UNKNOWN");
             }
             ret = GetQHYCCDReadModeResolution(m_CameraHandle, readModeInfo[rm].id, &readModeInfo[rm].subW,
-                &readModeInfo[rm].subH);
-            if (ret == QHYCCD_SUCCESS) {
+                                              &readModeInfo[rm].subH);
+            if (ret == QHYCCD_SUCCESS)
+            {
                 LOGF_INFO("Sensor resolution for mode %s: %dx%d px\n", readModeInfo[rm].label,
-                    readModeInfo[rm].subW, readModeInfo[rm].subH);
-            } else {
+                          readModeInfo[rm].subW, readModeInfo[rm].subH);
+            }
+            else
+            {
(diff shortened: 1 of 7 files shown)


indi-qhy/qhy_ccd.cpp
```diff
@@ -1195,6 +1193,8 @@ bool QHYCCD::setupParams()
 {
 
 
+    LOG_DEBUG("setup params\n");
```
knro, line 1196: This is not very useful log, please comment out. Also, remove any "/n" from other logs since it's not needed.

knro (changes requested):
Minor changes only. Did you test all combinations? Stream --> Capture --> Stream. Subframed streams...etc?

indi-qhy/qhy_ccd.cpp
nrackwit: Sheesh, agreed! Newlines are also removed. 
I wasn't too sure about the ones that are in IDLog(..) and left those in in the code. Let me know if those should go, too.

nrackwit:
For a short note on what I tested and what works and what doesn't:

Successfully tested:
a) Successive stream capture: i.e. connect camera, start streaming, start recording, stop recording, start recording, ..., stop streaming,
b) Switching between stream and exposure modes: i.e. capture exposure, start streaming,  record frames, stop streaming, capture another exposure,...
c) Modify frame resolution and origin offset via the "Image settings" tab,
d) Modify binning via "Image settings" and capture stream,
e) Capture stream in high resolution read mode (QHY294M).

What isn't working:
f) Frame resolution and origin offset for stream capture can't be modified via the properties in the "Streaming" tab, those won't take effect. 
g) In some cases the QYH294M would refuse to start streaming. That is, it would get into streaming mode but would not return the first frame. In this case the streamer object will stay in busy state and would not allow ECOS to go out of 'streaming mode' (Error "Cannot take exposure while streaming/recording is active."). If that happens only disconnecting the camera and restarting Indi seems to bring things back. I couldn't find a way to handle this in the driver code but also didn't look into the streamer code yet.

About the QHY294M not going into streaming seems to be related to the camera itself needing certain frame resolutions for streaming to work. I.e. any 1:1, 16:9 or 4:3 would work well, even at the maximum supported frame size for that format, but streaming at the native frame format would not work. I'm going to have to ask about that in the QHY forum. For what it's worth it also happens the same when I run the streaming test app that comes with their SDK. 

For the QHY294M to correctly work in streaming mode and to prevent the camera from stalling it is important to have the USB traffic value at a very low number. When streaming at 'zero' USB traffic value frame rates are actually quite decent, also for the larger resolutions (go figure, I finally learned that the the slider needs to go to the lowest value to get the highest streaming rates).

knro:
Thanks, I will share your feedback with QHY.

PR #349, opened 2021-03-09
indi-asi-power Add DSLR control

ken-self:
Control the DSLR shutter port on GPIO 21. Allows setting exposure, count and delay. Accurate timing to within about 1ms or so.
Arbitrary limit on exposures to 3600s.

### DIFF
--- a/debian/indi-asi-power/changelog
+++ b/debian/indi-asi-power/changelog
@@ -1,9 +1,14 @@
+indi-asi-power (0.91) buster; urgency=medium
+
+  * Add DSLR port control
+
+ -- Ken Self <ken.kgself@gmail.com>  Tue, 09 Mar 2021 12:20:00 +1100
+
 indi-asi-power (0.9) buster; urgency=medium
 
   * Initialize device type to None
   * Fix hardware revision and pigpio version reporting
   * make pigpiod a simple non-forking service
-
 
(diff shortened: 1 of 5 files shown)


knro:
@ken-self Thank you for your work on this driver! Do you think you can work on a generic INDI GPIO driver using the same code?

ken-self:
Yes, definitely. That was the next stage of the plan :)
So now we have 3 modes of operation: on/off, PWM and timed (long) pulses and it is relatively simple to enable those to be assigned to any GPIO pin.
Just one minor issue has cropped up with the package. Raspbian includes a package (not in standard Debian distro) which installs the pigpiod service  also and creates a conflict. This affects Astroberry users as it is Raspbian based. I can either suggest a work around with a dpkg --force-all or include a Replaces option in the control file. Any preference?

knro:
Replaces option might be more convenient as end users are not required to take any further actions... but is this OK? no functionality lost by replacing the other package?

Looking forward to the GPIO driver! Very excited for it finally!

ken-self:
No problem replacing pigpiod. Its just the service file and the daemon binary which I only build because it is not in the main debian repository. Another option is to add the raspbian repository to sources.list to get the pigpiod package and exclude it from our driver package. http://archive.raspberrypi.org/debian/pool/main/p/pigpio/
But I imagine that could open up a whole lot of other problems.
I've tested the Replaces option and it works fine. Raising a PR

PR #355, opened 2021-03-14
v0.1 Raspberry Pi GPIO driver

ken-self:
Initial Raspberry Pi GPIO driver.
Features: 5 ports assignable to any valid GPIO (determined by RPi hardware type)
On/Off, PWM and timed on/off on any of the 5 ports
Currently rebuilds and installs the same pigpiod service also used in indi-asi-power.


knro:
Maybe would be a good idea now that separate the pigpiod files into their own 3rd party library or something?

ken-self:
Ok. I'll start working on that.

ken-self:
Created libpigpiod to install daemon and systemd service. Also install pigpio.h needed to compile drivers
Removed equivalent from indi-asi-power and indi-rpi-gpio.

knro:
You're awesome!! Thank you for this great work. Do you think you can write some documentation for the GPIO driver to include in indilib.org device documentation?

ken-self:
No problem. I'll start writing it up

PR #361, opened 2021-03-20
usbfs_memory_mb: revert to setting 256MB as system defaults are too small

xsnrg:
@knro This is the fix/revert for the conversation around the ASI camera not able to capture in the forum.

### DIFF
--- a/libasi/99-asi.rules
+++ b/libasi/99-asi.rules
@@ -1,2 +1,3 @@
+ACTION=="add", ATTR{idVendor}=="03c3", RUN+="/bin/sh -c '/bin/echo 256 >/sys/module/usbcore/parameters/usbfs_memory_mb'"
 # All ASI Cameras and filter wheels
-SUBSYSTEMS=="usb", ATTR{idVendor}=="03c3", MODE="0666" 
+SUBSYSTEMS=="usb", ATTR{idVendor}=="03c3", MODE="0666"


knro:
This was removed specifically so it doesn't pollute any system-wide settings. With this change INDI returns to overriding usbfs_memory_mb. What if the system admin set that value to 512mb? then this would bring it down to 256mb.

xsnrg:
Thanks for the description, perhaps it should check if less than 200 then set it?

knro:
I think this should be approached differently. Each package was setting these numbers before. Maybe [postinst](https://github.com/indilib/indi/blob/master/debian/postinst) script in INDI (not 3rd party). Check if < 256, if yes, then set to 256. That would work. A similar approach could be made for RPM packages?

xsnrg:
If we set it at the INDI layer, it could cover all drivers, and we can assume that if they are using INDI, the user is likely using some form of USB camera, so this is not a bad idea. I only saw this in the ASI driver, were similar entries removed for QHY, etc?

xsnrg:
I gave this some more thought.

A post-install script would have to do some things to make permanent changes to the OS that we probably don't want to mess with, and if we don't make those changes, the setting would be lost at next boot. The OS level configuratios needed would be settings like kernel boot flags, or creating a start-up script for the OS. These are things that can get lost or overwritten easily with updates as well, and could end up wasting memory on the system if the devices are changed out and need lower requirements in the future. 

We want this setting to apply at every restart, or actually at every device attachment. This is where putting the setting in a device udev rule was actually rather clever. If we change the script in the udev rule to be like the code block below, we could leave the memory setting with the device driver code that needs the memory, instead of assuming that if you are running INDI, we will set the memory globally.

`if [[ $(cat /sys/module/usbcore/parameters/usbfs_memory_mb) -lt 200 ]]; then     $(/bin/sh -c '/bin/echo 256 >/sys/module/usbcore/parameters/usbfs_memory_mb');     else     exit 0;  fi`

The nice thing about doing it this way means that every driver that is loaded, that needs more memory, will adjust the setting higher, but only as high as is required by the driver that needs the most memory. A system that is not using drivers with big memory needs for USB will use a smaller setting. This also keeps it dynamic based on the USB devices plugged in. The above script is small enough that we can tuck it into the udev rule file and not have to create another file, or we could bundle it with INDI, and just call the script by name from any INDI driver that needs it. Putting the script in INDI means we only have to manage it in one place, and the snippet in the udev rule can just pass the numerical memory need as a parameter. 

`sh -c '/PATH/TO/INDI/script 256'` - for a driver that needs usb memory set to 256.

I don't like having the script in INDILIB as much because we create a dependency between two packages, the driver for the quantity, and INDILIB for the script. I would rather the driver be more self-contained, and the workings for the configuration understandable by looking at a single git repo, so I ruled this out, and suggest that each driver library has its own udev script if it needs one.

A scenario to show how this would work:

A system has 3 USB devices, so 3 drivers, A, B, and C. 

A requires 100, B requires 300 and C requires 200.

A is connected first, sees the system is set to 16, changes the setting to 100 through udev. A is happy.
B is connected next, sees the system is set to 100, and bumps it to 300. B is happy, but so is A.
C is connected next, sees the system is already set to 300, so does not change the setting. C is happy, B is happy, A is happy.
If the system is already set to 300 or higher by something external to INDI, nothing changes and A, B, and C are already happy.

If the system at next boot only connects A then the system runs with that device and an allocation of 100 instead of 300.

This also fixes the problem of lowering a value that may have been set outside of INDI drivers (sysadmin, other applications, etc). We would never lower the value. We would only bump be bumping it up if needed.

Thoughts?

xsnrg:
https://www.flir.com/support-center/iis/machine-vision/application-note/understanding-usbfs-on-linux/

This is also interesting, and shows that this setting is used  by the areas that @pawel-soja is currently working on with buffers. Multiple cameras capturing at the same time is another consideration. What if I have a 6200 and guide with a 1600... sure, not likely, but a 294 with a 290 or 224 happens all the time. Main capture camera and guider. The most stable configuration might be a udev rule that adds ALL of the memory it needs to the existing setting, so that each device can guarantee it will get the buffer needed to run at full speed and not have USB memory contention with another device. It should then remove that same amount of memory from the setting when unplugged. There are limits to this that need to be accounted for, of course.

knro:
There is a simpler solution, which is to add this to indi-auxiliary.rules file. I'll do that.

xsnrg:
It sure would be nice if a device could just tell the kernel how much memory it needed and it would be allocated to it if available. Maybe someday that will be in the kernel. For now a setting of 256 in indi-auxiliary.rules should cover many cases, and as you said, it is simple.

PR #384, opened 2021-04-25
Fix warnings - Add build for MacOS

pawel-soja:
Fixed warnings that appeared when compiling on MacOS.

The big change to cmake is the new InstallImported module.

Example:
```cmake

add_library (ASICamera2  SHARED IMPORTED)
add_library (EFWFilter   SHARED IMPORTED)
# ...

set_target_properties (ASICamera2  PROPERTIES VERSION 1.16 SOVERSION 1)
set_target_properties (EFWFilter   PROPERTIES VERSION 1.6 SOVERSION 1)
# ...

set_property (TARGET ASICamera2  PROPERTY IMPORTED_LOCATION "some_arch/libASICamera2.bin")
set_property (TARGET EFWFilter   PROPERTY IMPORTED_LOCATION "some_arch/libEFWFilter.bin")

# ...

# Install library - NEW FUNCTION below
install_imported (TARGETS ASICamera2 EFWFilter DESTINATION lib)

```
Activities performed in stages:

- Configuration: none
- Build
  1. Copy the files from the source to the current build directory with a diffrent name -  prefix + target + suffix + version
  2. Create symblic links file_without_version -> file_with_soversion -> file_with_version
- Install
  1. Copy library and symlinks to system directory

In short, the same behavior as building a library from source.

There is order in the build directory.
image

image


knro:
Thanks, we'd have to check how this affects Debian packages as well.

pawel-soja:
@knro When are you able to check it?
Conflicts in the project start to arise.

pawel-soja:
@knro Please consider testing and adding this PR before accepting new drivers (like Bresser Exos 2 GoTo).

knro:
@pawel-soja Sure, will have to check also with @xsnrg that this does not break RPM-based builds.

knro:
Darn it, I accidentally merged this (again). Ok, this definitely breaks debian files and installs the libraries to /usr/lib and not /usr/lib/arch (e.g. /usr/lib/x86_64-linux-gnu/). Sorry there is no "un-merge" so it has to be re-submitted again :(

pawel-soja:
Not enough coffee :)
You're right! I lost CMAKE_INSTALL_LIBDIR.
Fixes are already on the branch. I'll throw PR.

PR #393, opened 2021-05-04
Replace pigpio timer with INDI timer

ken-self:
Add -m option when starting pigpiod to reduce idle CPU usage
Replace the pigpio timer in indi-asi-power and indi-rpi-gpio with the built in INDI timer as the pigpio timer does not work with the -m option.

### DIFF
--- a/indi-asi-power/asipower.cpp
+++ b/indi-asi-power/asipower.cpp
@@ -21,11 +21,13 @@
 #include <string.h>
 #include <math.h>
 #include <config.h>
+#include <chrono>
 #include <pigpiod_if2.h>
 #include <asipower.h>
 
 static class Loader
 {
+public:
     std::unique_ptr<IndiAsiPower> device;
 public:
     Loader()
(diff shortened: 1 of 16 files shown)


indi-asi-power/asipower.cpp
```diff
@@ -34,19 +36,14 @@ static class Loader
...
-
 IndiAsiPower::IndiAsiPower()
 {
     setVersion(VERSION_MAJOR,VERSION_MINOR);
     std::fill_n(m_type, n_dev_type, 0);
-    dslr_cb = 0;
-    dslr_end = dslr_last = 0;
     dslr_isexp = false;
     dslr_counter = 0;
+    timer.callOnTimeout([this](){IndiTimerCallback();});
```
pawel-soja, line 45: You can use simply:
```cpp
timer.callOnTimeout(std::bind(&IndiAsiPower::IndiTimerCallback, this));
```

pawel-soja: Examples: https://github.com/indilib/indi/pull/1386#issuecomment-830839932

PR #398, opened 2021-05-11
Weather Radio V 1.13 | Rain sensor, time lapse optimization, HTML redesign

sterne-jaeger:
* Rain sensor integrated
  * Better handling of long exposure times for timelapse
  * Improvements for auto exposure during dawn/dusk
  * Pan tilt controller support added
  * Redesign of the HTML page
  * Firmware optimization and bug fixes
  * Bug fixes for the INDI weather radio service


knro:
Thank you Wolfgang! This is becoming better & better! I really think this project deserves to be featured here: https://indilib.org/weather.html

Do you think we can copy the documentation to there? a lot of user end up on the devices pages to see what's working and what's not.

sterne-jaeger:
Many thanks for your kind words!
I already thought whether it would make sense to extract it from indi-duino and make it a standalone 3rd party package. And yes, adding it to the top level documentation is a better place for an entry point documentation.

PR #402, opened 2021-05-15
add Vixen Starbook Ten driver

bwalex:
This is a (lightly tested) driver for Vixen Starbook Ten-based mounts, with support for pulse guiding, etc.


knro:
Thank you for the driver submission! A few questions:

1. What distinguishes this driver from the regular Starbook?
2. Can you please also add the debian files for it? You can duplicate the existing debian/indi-starbook directory and update accordingly.
3. Why httplib? can this be done via libcurl?

bwalex:
Hey there!

RE 1 - Starbook and Starbook Ten are pretty different in what they support, and how some of the APIs work. In theory one could support both in one driver, but that ends up being more overhead than just having a different driver for what really is a different device.

RE 2 - I can look into that.

RE 3 - Two main reasons - I'm not terribly fond of libcurl for something like this - it has a lot of baggage for a very simple use case as this one, and it also has more latency/overhead than this http library; and, second, I'm also hoping that we can eventually upstream this into indilib/indi, where I don't think new external dependencies (such as libcurl) are welcome. Personally I'm also not terribly fond of unnecessary external dependencies.

bwalex:
Alright, I've done my best with the debian packaging bits (same really goes for what I assume is the rpmspec file) - I use ArchLinux, so am not really familiar with either of them :)

knro:
Thanks! That leaves documentation as the last requirement. We need to add documentation to INDI website.  Perhaps similar to the existing INDI [Starbook driver documentation](https://indilib.org/telescopes/vixen/vixen-starbook.html). If you have the content, can you upload everything including the screenshots in a zip file here so I can update the website?

bwalex:
Thanks for getting that merged! I'll have a look at providing docs within a week or so.

PR #403, opened 2021-05-16
Updates to ZWO SDKs

jpaana:
ZWO updated almost all of their SDKs, unfortunately changelogs are just "Fix some bugs":
ASI Camera SDK 1.18
EAF SDK 1.4
EFW SDK 1.7

I could only test ASI178MC, ASI120MM-S and EAF and those worked fine, but can't say if the issues with 1.17 are fixed or not as that also worked fine for me.


knro:
Thanks, I have the same equipment as yours. But I think we need more testing for this before we merge again.

knro:
They work OK here. @pawel-soja Any issues with them?

pawel-soja:
ASI178MM, ASI120MM-Mini, ASI120MC-S (USB 3.0 :) ) works
I remember there was a problem with the last update, but with a bigger cam.

PR #420, opened 2021-06-28
Add meadecam driver for x86 and x86-64

bgilsrud:
This commit adds the meadecam driver which is a driver for Meade's rebranded
Touptek cameras. This has been tested with an LPI-GM and LPI-GM Advanced
camera. I suspect it also supports the DSI IV cameras.

I was able to get frames from the LPI-GM and LPI-GM Advanced using Ekos
on an x86-64 machine.
I also verified that this doesn't break the build on an arm64 machine.


knro:
Thanks! what about ARM 32bit and 64bit?

bgilsrud:
Unfortunately, Meade doesn't provide support for ARM and they've indicated that they have no plans to do so. That's a real bummer since I would like to use my cameras on an ARM SBC. I'm planning to eventually write a new driver to work around this.

openastroproject:
> Unfortunately, Meade doesn't provide support for ARM and they've indicated that they have no plans to do so. That's a real bummer since I would like to use my cameras on an ARM SBC. I'm planning to eventually write a new driver to work around this.

That seems a strange attitude given that ARM environments are supported by many other rebranded Touptek cameras.

However, I bet if one took the ARM library for, say, Touptek's own cameras, and patched in the relevant USB VID and PID values and camera names from the Intel Meade SDK, then it might well work.  It may have to be done at runtime using a library loaded using dlopen(), but it's not particularly hard.

bgilsrud:
I agree. I even mentioned that other vendors (Orion, Altair, ...) provided ARM libraries and that apparently wasn't convincing to them. Meade doesn't even provide an SDK :(. These libraries are plucked from the SkyCapture software. It's almost like they don't want people to use their cameras.

I agree with most of what you're suggesting, but it could be a little trickier than that. If it's just the PID/VID that's one thing, but i think there may be more complicated structures that are used to define the camera in the library rather than discovering it from the device.

openastroproject:
I've done something fairly similar with the Altair SDK because they no longer support some of their earliest rebadged Touptek cameras.  As long as the cameras are functionally identical and the matching PIDs and VIDs are replaced (there's a table of them in the library) then it may well work.  And it's much faster to try and to test than reverse-engineering the library to write an open source driver (been there, done that, with some of the now-unsupported QHY cameras), even if having an open source driver is the more desirable long-term goal.

bgilsrud:
@openastroproject can you point me at code that does what you're suggesting?

openastroproject:
This might help: https://github.com/openastroproject/openastro/blob/master/liboacam/touptek/dynloader.c

It loads the dynamic library at runtime using dlopen() rather than using the linker at compile time (and uses a version of the library that I have renamed).

From memory (which may be faulty), the code has a table of PIDs for the cameras, grouped together if they use the same internal interface.  If you look at the _patchLibrary() function, it takes the address of the Toupcam_EnumV2 function in an old release of the toupcam library, calculates the address of the table of PIDs from data in the code and patches a new (or rather, old) PID into the end of the table to allow the library to support the old Altair MT9M034M GP-CAM.  It's not quite the same as I'm suggesting you could do here, and it's x64-based rather than ARM, but it's reasonably close to what I'd try for the Meade cameras -- find the PID table, find the PID of the Toupcam camera that is rebadged for Meade and replace it with, or add, the Meade PID, assuming of course that the PID is even different.

There's probably (actually there is, I've seen it)  somewhere in the libtoupcam code that compares the VID with 0x0547 for Toupcam.  That value would also need patching to 0x0548(?) for Meade.

Not particularly related to this, the one possible difficulty with support for the Meade cameras is that the licence for the Meade software may prohibit distribution of parts of the application (specifically the libraries in this case) by a third party.  I haven't checked as I don't have it.

bgilsrud:
The meade software has no license.

knro:
If this is Touptek based then Meade *should* have arm32/arm64 binaries as well.

bgilsrud:
I figured out how we can provide our own camera info to the toupcam library. This should allow us to use any OEM touptek camera (Meade, Altiar, Orion, RisingCam, ...) with the official touptek sdk. I'm going to close this PR and I'll work on implementing this.

PR #424, opened 2021-07-12
Update weatherradio.cpp

minhlead:
Increase sky temperature coefficient limit to adapt the driver for tropical region. The current K2 coefficient (-50 to 50) is not enough for temperate region where ambient temperature higher than 40 degree.


sterne-jaeger (approved):
Makes sense.

PR #429, opened 2021-07-26
libindi-dev added to the required packages list

thx8411:
Hi,

I just re-installed everything from scratch on a new laptop. In my mind, "libindi-dev" should be added to the required packages list, in the build section of the README.md file.

Best regards !

Thx8411


knro:
So there is one issue here. If depends whether you are building INDI 3rd party library against a compiled version of libindi or not. If against a compiled version, then no need to override that from the system repos. So perhaps we put this above and indicate this to avoid confusion?

thx8411:
Hello,

The "Install Pre-requisites" section mentions these packages : 

libgps-dev
libavcodec-dev
libavdevice-dev

missing in the "Building individual 3rd Party Drivers (typically recommended)" section.

Should we add them ?

Best regards,

Thx8411

knro:
Right, for *complete* installation those are needed. Though what uses libgps-dev now?

PR #444, opened 2021-08-31
Toupcam OEM camera support

bgilsrud:
This adds support for the Meade cameras that are produced by ToupTek to the toupcam driver.


knro:
I forgot about the XML file, shouldn't we add an entry there?

PR #445, opened 2021-08-31
Work on SBony SV305M Pro camera support

thx8411:
Hi !

Some light refactoring on the cameras capabilities handling, in order to add the SV305M Pro support.
The SV305M Pro should work, but not tested yet (I don't own the device).

Best regards !

Thx8411


knro:
Thanks. Can we work on the documentation? I need a jpg image of each tab in the driver. And perhaps some description on any special features. Anything to alert to user to? We really need to add proper driver documentation. You can zip the information and attach it here.

thx8411:
Hello Jasem,

Sure, I will work on it. Should I use this kind of page as a template ? : 

https://www.indilib.org/ccds/player-one-astronomy-cameras.html

Best regards,

Blaise

knro:
Sure that would work! Add as much detail as you believe necessary, including any quirks or warnings.

knro:
I'll merge this for now, but please let me know whenever you have the documentation ready as it is very critical to have at this stage with increasing number of users for this driver.

PR #454, opened 2021-09-09
Adopt ASI driver to simulator behavior and like QHY driver

mworion:
Hi Jasem,

based on our discussion on INDI forum, I cleaned up the VERBOSE_EXPOSURE time as constant and adjusted the sending of CCD_EXPOSURE = 0 only if the exposure is finished. The solution is used in the QHY driver as well, which also only reports CCD_EXPOSURE back if timeLeft is bigger than 0.0049 s. Both driver set the CCD_EPOSURE value to 0 explicit anyway if exposure is finished.
I cannot build myself, so please review an accept if ok or comment if wrong.

Michel


knro:
Looks good, thanks!

PR #462, opened 2021-09-21
SV305 : Mac OS X 64 support added

thx8411:
Hello,

This PR should add Mac OS X 64 bits support to the SV305 driver.
I don't have any way to check it. Could a Mac expert have a look on it ?

Thanks a lot !

Best regards,

Thx8411


knro:
Looks good thanks! Let's merge and find out actual reports from end-users.

Issue #464, opened 2021-09-27
 indi-gige does not compile with current version of Aravis


Jul3k:
The current implementation of  indi-gige requires to be compiled with Aravis < 0.8.0. as function calls were changed to include an [error parameter](https://github.com/AravisProject/aravis/commit/36d23d9a3f3e6e4118e3f8e80c5bfed000632d0d). I do not know how mature this driver is. I was trying to get a USB3 Vision camera working which is supported by Aravis.

**Reproduce**
1. Install Aravis 0.8.18
2. Follow build instructions
3. Compile error:
```
    indi-3rdparty-1.9.1/indi-gige/src/ArvGeneric.cpp:198:53: error: too few arguments to function ‘const char* arv_camera_get_device_id(ArvCamera*, GError**)’
    this->cam.device_id   = arv_camera_get_device_id(camera);
                            ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```

**Expected behavior**
Should compile with current version of Aravis.

**Desktop:**
 - Arch Linux

knro:
I don't think driver is currently well maintained. Can you please submit a patch to fix the issue?

Jul3k:
The driver compiles correctly with the submitted patch. It does however not show in Ekos because it says no camera is connected:
`"Unable to establish:\n+ GigE CCD\nPlease ensure the device is connected and powered on."`
This is strange because I can see in the Debugger, that a camera is found and initialized. Do you know where this error message comes from? I cannot find anything similar in the source code.

knro:
You can safely ignore this if the devices indeed shows up in Ekos. What's the name of the device that shows up exactly?

Jul3k:
The problem is that no device tab shows up in the Indi Control Panel. I can select a `GigE CCD` in the profile editor, but when I start the server, no new device tab is shown in the indi control panel. The log shows the following:

`[2021-09-28T14:30:14.801 CEST DEBG ][           org.kde.kstars.indi] - Received new device U3-380xCP-C
[2021-09-28T14:30:14.806 CEST DEBG ][           org.kde.kstars.indi] - < U3-380xCP-C >: < CONNECTION >
[2021-09-28T14:30:14.806 CEST DEBG ][           org.kde.kstars.indi] - < U3-380xCP-C >: < DRIVER_INFO >
[...]

PR #465, opened 2021-09-28
Updated indi-gige to compile with aravis 0.8

Jul3k:
The current version of Aravis introduced an addition error parameter in its function calls. This parameter has been added and the CMake dependencies updated to aravis 0.8. See #464

### DIFF
--- a/cmake_modules/FindARAVIS.cmake
+++ b/cmake_modules/FindARAVIS.cmake
@@ -19,15 +19,15 @@
 
 SET(ARV_FIND_REQUIRED ${Arv_FIND_REQUIRED})
 
-find_path(ARV_INCLUDE_DIR aravis-0.6/arv.h)
+find_path(ARV_INCLUDE_DIR aravis-0.8/arv.h)
 mark_as_advanced(ARV_INCLUDE_DIR)
 
-set(ARV_NAMES ${ARV_NAMES} aravis-0.6)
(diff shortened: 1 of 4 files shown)


knro (approved):

rkaczorek:
aravis 0.8 is not available for raspbian buster (old stable) so this commit breaks build dependencies for indi-gige

knro:
Good point, I didn't catch this. Any workarounds possible?

Jul3k:
Still after the update to Aravis 0.8 I was not able to have indi-gige properly running with the device I could test with. It is hard for me to comment on the overall functionality of the driver though.

Jul3k:
Well, if it is reverted it will not compile on systems that have aravis 0.8, right?

rkaczorek:
@Jul3k does it mean that aravis 0.8 is not backward compatible? If this is the case, could you verify if aravis version detection is possible at compilation and adjust submit PR to handle this?

Jul3k:
Yes aravis 0.8 introduced a change to the API that broke backward compatibility. I think a compile time check based on defines for the used aravis version is possible but I can't take care of it for the moment. While I believe that a driver that can access gige and genicam standardized cameras is very useful, I am a bit skeptical about the general functionality of the driver in its current state. The code contains some functions that are generic to BlackFly cameras and it seems that some functions which are useful for astrophotography are not standardized and require vendor specific handling in general. I was working on implementing IDS cameras but finally gave up.

PR #472, opened 2021-10-12
xc_cleanup: change the library and driver spec files to match

xsnrg:
Note that the library still uses underscore while other places use dashes.

The spec files were left behind when the folder names and files changed a while back.


knro:
@iliaplatone can you please check?

@xsnrg Looks like failing on MacOS?

xsnrg:
> @xsnrg Looks like failing on MacOS?

yes it does appear to fail on MacOS, but nothing I changed would have affected that, so it was likely failing before. These are only spec file changes for Fedora.

PR #488, opened 2021-11-17
Update qhy_ccd.cpp

sonny486:
Found the issue with my QHY 600m showing up with a BAYER pattern.

Looked at the original code and it seems it wasn't complete.

I edited the code and cleaned it up a bit.

I have tested with both my QHY600m and it now shows up as a mono camera, and I have tested with my QHY247c and it shows up as color with proper BAYER pattern now.

@knro please review when you can.


sonny486:
@knro Please reveiw.  Thanks.

PR #490, opened 2021-11-20
SV305 : live view subframing and binning enabled

thx8411:
Hello,

Subframing and software binning is now available in live view.
Up to 260 fps can be achieved on USB 3.0 cameras (SV305 PRO and SV305M PRO). 75 fps on the USB 2.0 SV305 camera.

Best regards !

Thx8411


knro:
That's awesome!! Curious to see what the numbers are on a Raspberry PI4 if you happen to test those?

PR #492, opened 2021-11-24
Update ASICamera to 1.20.3

paolostivanin:
This fixes a segfault on some systems (e.g. openSUSE Leap 15.3) introduced by version 1.20.2.

Verified on openSUSE Leap 15.3 and Tumbleweed.


knro:
What's the reason behind the header file change? did you run dos2unix against it?

paolostivanin:
Nope, I forgot do that! Just done now, no diff!

PR #503, opened 2021-12-17
move udev rules from /etc/udev/rules.d to /lib/udev/rules.d

alteholz:
Some time ago (with version 0.140-1 of the udev package) the handling of udev rules was changed. The default rules shall now live in /lib/udev/rules.d.

Most indi drivers already comply with this, except indi-dsi and indi-orion-ssg3 (those are updated with this PR).
Other indi driver just need an update of their documentation.

More information about this can be found in [1].

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=586668


knro:
Thanks!! @xsnrg Is this OK with you as well?

xsnrg:
yes, this should be fine.

knro:
@alteholz I take it we need a similar PR in INDI core as well?

alteholz:
@knro oh, yes, if nobody beats me, I will have a look

alteholz:
@knro INDI core already moved to /lib/udev ...

PR #507, opened 2021-12-26
Enable binning on gphoto

thx8411:
Hello !

I have a suggestion : enabling software binning on gphoto devices.

I mostly use my DSLRs with Stellarmate OS and wifi.

With recents high resolution sensors, downloading RAW subframes takes ages.
Binning helps a lot, especially when focusing.

What do you think ?

Best regards !

Thx8411


knro:
How do you bin bayer frames? RAW images (e.g. CR2) ...etc?

thx8411:
Hi Jasem,

Binning only applies on fits images, not "native".
But you're right, odd binning factors should be avoided.
Even binning factors result in a kind of "mono sensor", with a bad QE and mostly centered on green spectral response.

Should I remove the "CCD_HAS_BAYER" capability when binning ?

Best regards,

Thx8411

knro:
Binning was considered before and was dropped for this reason. You cannot properly bin a bayered frame. The binning function only works for mono frames which we don't have in this driver.

thx8411:
I agree. Binning a CMOS color sensor doesn't have sens for imaging.
But for previews (focusing, plate solving, etc.), it reduces data flows by 4 (bin 2), and thus the time between each frame by the same factor.

On my configuration : INDI drivers on a Stellarmate OS RPI 3+, kstars on my laptop, using wifi, downloading a canon EOS M2 shot takes 8s.

Best regards,

Thx8411

knro:
So it's binning the bayer frame? that doesn't produce bad/unusable images? One possible way to do this is to take the green component of the image, bin that, and then send it over as Mono. That would be sufficient.

jochym:
Maybe use in-camera binning by changing RAW resolution? Many cameras have sRAW and multiple raw resolutions. Or if it  is for focussing/framing, use JPEG. Using RAW for this is really suboptimal (if not outright unwise). Perhaps we should make it less manual? Like switch with setting for resolution used in framing/focusing.

knro:
RIght, many cameras have "Small jpeg" type of format as well. Though this wouldn't be binning per se since it's not going to end up 1/2 or 1/4 of the size. Maybe 2x2 --> Medium Jpeg, 4x4 --> Small Jpeg.

thx8411:
Right, good idea !

First, I will limit binning to even factors
Second, I will work on a dedicated binning function, instead of using the default one.

Like you said, extracting only the green layer is a good idea.
We could also apply some RGB to luminance correction (but finding the proper colospace could be a bit tricky), and it consumes CPU.

Best regards,

Thx8411

jochym:
Yes, it is often 2/3, and 1/2. My EOS 40D is: 3888 x 2592 and other: 2816 x 1880, 1936 x 1288 which is close to these rations. I think it does some cropping and then binning, since aspects are different in each case. Furthermore, using in-camera binning makes it faster since it is hardware-driven (DSP in the camera).

jochym:
Although, as far as I know sRAW and other small-resolution data in Canon and Nikon cameras are not binned but down-sampled data - so it is good for framing/focusing but should not be thought of as real RAW data. Furthermore, at least for Nikon it seems to be in different color space and has some tone curve applied. For Canon, it seems to be some form of kernel sub-sampling with 4x2 pixels kernel and bit-packing to compress the data.
Still, it seems to be a good way to get fast refresh on low-power devices and/or slow links (WiFi). I think that for framing/focusing, high quality JPEG may be enough.

thx8411:
Thanks for you comments !

Wouldn't jpeg introduce artifacts which could disturb FWHM or plate solving ?

Indeed, there is many options. I will work on it.

Should I close the PR, or let it open for further discussions ?

Best regards !

Thx8411

jochym:
If you need more technical analysis of sRAW formats, here is a paper I stumbled upon: [sRaw format analysis by Douglas A. Kerr](http://dougkerr.net/Pumpkin/articles/sRaw.pdf)

jochym:
As for plate solving - I doubt it @thx8411. The centroid should not move in any substantial way. Focusing may be the only case - I do not think it will be influenced with high-quality JPEGs. However, this is only *my guess*. The artefacts will probably increase measured FWHM but should not shift the *position* of the minimum because the general effect of JPEG is reduction of spatial resolution, and it applies uniformly to focused and unfocused images. Only experimentation can tell us how strong this effect is.

knro:
I think it's OK for both cases as long as it is *consistent*

thx8411:
Hi folks,

After some thinking, binning a bayer frame is pretty easy.

Let see 2x2 binning as a kind of convolution filter you apply on one of two columns and one of two rows.

A 2x2 binning matrix would be

0 0 0
0 1 1
0 1 1

On a bayer frame, if you consider each 2x2 bloc of pixels as a "meta pixel", if you apply this matrix

0 0 0 0 0
0 0 0 0 0
0 0 1 0 1
0 0 0 0 0
0 0 1 0 1

on all the pixels of a "meta pixel", but just on one of two "meta pixel" columns and one of two meta pixel rows, you get
a binned bayer frame.
You don't even have to care of the bayer pattern.

You just have to be sure you have a %4 width and high frame size.

Obviously, I won't use convolution filters. It's just the simplest way I found to explain my idea.

What do you think ?

Best regards,

Thx8411

knro:
That's actually a pretty **smart** idea, if I understand it correctly! How expensive is this operation?

thx8411:
Hi Jasem,

"How expensive is this operation?"

In my mind, not much more than usual binning. Principles stay the same, it's just a matter of x and y offsets.

I will work on an implementation, and come back to you.

Best regards

Thx8411

thx8411:
Hello !

This is a first try in 8 bits :

`
uint32_t BinFrameOffset;
            uint16_t val;

            for (uint32_t i = 0; i < SubH; i++)
                for (uint32_t j = 0; j < SubW; j++)
                {
                    BinFrameOffset=(((i/BinX)/2)*2+i%2)*SubW/BinX+((j/BinX)/2)*2+j%2;
                    val=BinFrame[BinFrameOffset];
                    val+=(RawFrame[i*SubW+j]/(BinX*BinX));
                    if(val>UINT8_MAX)
                        val=UINT8_MAX;
                    BinFrame[BinFrameOffset]=(uint8_t)val;
                }
`
The idea was to browse the full frame, and look where each data should be put in the binned frame, instead of browsing the binned frame, looking for the data to catch in the full frame.

This solution looks elegant, but with two major drawbacks :

- it's damn slow (but there is plenty of room for optimization, I wanted the code to stay readable at this stage)
- the pixel averaging method leads to dynamic loss

But I get decent pictures...

Still working on it.

Best regards !

Thx8411

thx8411:
Some of my tests (8 bits and 16 bits).

Without any optimization, a 1280x1024x8 2x2 bayer binning takes about 29ms on my laptop (about 4ms for legacy binning). 

[test_16bits_bin_bayer.zip](https://github.com/indilib/indi-3rdparty/files/7798989/test_16bits_bin_bayer.zip)
[test_8bits_bin_bayer.zip](https://github.com/indilib/indi-3rdparty/files/7798990/test_8bits_bin_bayer.zip)

thx8411:
With this simple optimizations :

`uint32_t BinFrameOffset;
            uint16_t val;
            uint32_t BinW=SubW/BinX;
            uint16_t BinFactor=BinX*BinX;
            uint32_t RawOffsetH;
            uint32_t BinOffsetH;

            for (uint32_t i = 0; i < SubH; i++)
            {
                RawOffsetH=i*SubW;
                BinOffsetH=(((i/BinX)/2)*2+i%2)*BinW;
                for (uint32_t j = 0; j < SubW; j++)
                {
                    BinFrameOffset=BinOffsetH+((j/BinX)/2)*2+j%2;
                    val=BinFrame[BinFrameOffset];
                    val+=(RawFrame[RawOffsetH+j]/BinFactor);
                    if(val>UINT8_MAX)
                        val=UINT8_MAX;
                    BinFrame[BinFrameOffset]=(uint8_t)val;
                }
            }
`
Time reduced from 29ms to 15ms

knro:
The images you attached show a FITS Cube (RGB) and not a bayer image? RGB Cube binning is done for each channel separately and should not result in any issue. Do you have any binning done on a **bayer** image?

thx8411:
Hello Jasem,

Right, sorry. Debayer was enabled in my fitsviewer. So, files where recorder as FITS cubes :-(



Some minor code changes

`uint32_t BinFrameOffset;
            uint16_t val;
            uint32_t BinW=SubW/BinX;
            uint16_t BinFactor=BinX*BinX;
            uint32_t RawOffset=0;
            uint32_t BinOffsetH;

            for (uint32_t i = 0; i < SubH; i++)
            {
                BinOffsetH=(((i/BinX) & 0xFFFFFFFE) + (i & 0x00000001)) * BinW;
                for (uint32_t j = 0; j < SubW; j++)
                {
                    BinFrameOffset=BinOffsetH + ((j/BinX) & 0xFFFFFFFE) + (j & 0x00000001);
                    val=BinFrame[BinFrameOffset];
                    val+=RawFrame[RawOffset]/BinFactor;
                    if(val>UINT8_MAX)
                        val=UINT8_MAX;
                    BinFrame[BinFrameOffset]=(uint8_t)val;
                    RawOffset++;
                }
            }
`
Best regards

Thx8411

[8bits.zip](https://github.com/indilib/indi-3rdparty/files/7810328/8bits.zip)

[16bits.zip](https://github.com/indilib/indi-3rdparty/files/7810329/16bits.zip)

knro:
Outstanding! The results are great. How is this now performance wise?

thx8411:
Sadly, no performance improvement. gcc optimizers are pretty smart...

I will perform some serious perf measurements, to identify bottlenecks.

Best regards

thx8411:
Hello,

I pushed a proposal for bayer binning in the main indi repo (PR #1586).
I'm still playing with "perf" in order to get it a bit faster.

Best regards !

Thx8411

knro:
Bravo! merged! Can you please update this PR accordingly?

thx8411:
Hi,

It should be OK now. Tested on my DSLR without issues.

Thx8411

knro:
Bravo! Would this in any form affect live view? or how subframing?

knro:
Thanks! All looks good after I fixed a small issue in streaming. However, I found another issue not linked with this PR but maybe you can help track it down. It appears some threading issues are causing a SIGABORT when **recording**. After recording is stopped, I see two threads running with GPhoto thread waiting on some condition. Then in gphoto.cpp:1880 when it tries to join the liveViewThread, it aborts.

I think the issue is related to stopRecording getting issued in newFrame

PR #525, opened 2022-01-14
NexDome: Slave across Meridian fix

dscbiggs:
Fix to movement across 0 degrees that causes slaved motion to stall when dome crosses meridian.


knro:
Thanks! Please remove the .DS_Store file from the commit.

dscbiggs:
Removed .DS_Store and updated header to remove unused variable that gave compile error.

PR #528, opened 2022-01-17
celestronaux: add missing header file

hamarituc:
When compiling `indi-celestronaux` version 1.8.4 I got the following error message:

```
In file included from /var/tmp/portage/sci-libs/indilib-driver-celestronaux-1.9.4/work/indi-3rdparty-1.9.4/indi-celestronaux/auxproto.cpp:24:
/var/tmp/portage/sci-libs/indilib-driver-celestronaux-1.9.4/work/indi-3rdparty-1.9.4/indi-celestronaux/auxproto.h:120:9: error: ‘size_t’ does not name a type
  120 |         size_t dataSize() const
      |         ^~~~~~
```

This patch adds the necessary header file, so the type `size_t` is known to the compiler.

* Operating System: Linux
* Distribution: Gentoo
* Compiler: GCC 11.2.0

### DIFF
--- a/indi-celestronaux/auxproto.h
+++ b/indi-celestronaux/auxproto.h
@@ -24,6 +24,7 @@
 #pragma once
 
 #include <vector>
+#include <stddef.h>
 #include <stdint.h>
 
 typedef std::vector<uint8_t> AUXBuffer;


xsnrg:
This looks like it will fix the Fedora build as well:

/builddir/build/BUILD/indi-3rdparty-master/indi-celestronaux/auxproto.h:120:9: error: 'size_t' does not name a type
  120 |         size_t dataSize() const
      |         ^~~~~~

PR #534, opened 2022-01-29
indi-gphoto: Fix open() call

AdrianBunk:
S_IRUSR is a valid "mode" parameter for O_CREAT and O_TMPFILE.

S_IRUSR is not a valid "flags" parameter, and in addition to being
wrong it is architecture dependent which flag happens to have the
same value.

@knro This was introduced by commit aa3d611039728f4702ecbd477e35bcd39f3eb710.


AdrianBunk:
@alteholz This fixes the build failure on MIPS.

PR #536, opened 2022-01-31
Add libpigpiod, indi-asi-power and indi-rpi-gpio to CMakeLists 

ken-self:
@knro Please review as this is my first time modifying the main CMakeLists
Purpose is to add the drivers and library to the daily build. Previously the driver builds failed when the library code was split out.
Several end users have been asking about the drivers recently as they cannot find them on the PPA
Hope this resolves that

Tested by running a make on both arm and x64 devices to check that they build only on arm devices. Also ran "make install"" for the library and individal drivers on arm device and verified through Ekso.


knro:
Thanks, was this tested on an arm 32 and arm 64 bit systems that are NOT raspberry pi? Does it result in build failure?

ken-self:
Hi Jaem,
No it was tested on raspberry pi. I dont have access to other arm devices.

ken-self:
Out of curiosity I installed libpigpiod on x64 to see what happens. It fails (of course) because I think it is looking for /proc/device-tree which symlinks to /sys/firmware/devicetree/base but does not exist on that machine.
Also, looking into it a bit more outside my area of competence I understand that GPIO is embedded in the ARM architecture so I think libpigpiod could be usable on any ARM device.
So I think  it should be safe to install anwhere but it will only work on ARM devices where the device tree overlay is installed
But as I said this is outside my area of competence

knro:
Might use qemu to test this on Ubuntu 20.04 aarch64 for example to see if it fails.

ken-self:
I had a deeper look into pigpio.c which forms the basis of the libpigpiod. It looks like it assumes an ARM CPU in gpioHardwareRevision when it checks the hardware revision code which then determines which GPIO pins are available. 
I'm wondering if it is worth writing a wrapper script to check the CPU type for armv* or aarch64 using uname -p and also check /proc/cpuinfo for Hardware: BCM3825 or BCM2711. Those checks plus the existing revision check within libpigpiod should hopefully exclude any non-compliant hardware.

PR #537, opened 2022-02-04
Updated libahp-xc

iliaplatone:
Updated libahp-xc, still needs some adjustment into ahp-xc


k-ross:
Running CMake now prompts for a sudo password. This is undesirable. Just running cmake should not need root, nor should running cmake touch anything outside of the build folder.

k-ross:
Installing the libraries should be done from a Makefile, not from the CMakeFile.txt. CMake should generate the Makefile, and only when "sudo make install" is run, should the libraries be installed. 

Furthermore, I see you have *.deb files. This is also undesirable, since this limits support to only Debian-based distros. What about RedHat, Arch, CentOS, etc? INDI is available on those distros as well. The contents of the .deb files are just the library files, with no postinst or any other special requirements. Therefore, it would be easy to create a Makefile that just copies the libraries to the appropriate location, thereby allowing this to work on non-Debian distros.

PR #541, opened 2022-02-11
ASIAIR Plus power sensor support

tentaisan:
This patch support for the ASIAIR Plus power sensor for indi-asi-power.

### DIFF
--- a/indi-asi-power/asipower.cpp
+++ b/indi-asi-power/asipower.cpp
@@ -44,6 +44,10 @@ IndiAsiPower::IndiAsiPower()
     dslr_counter = 0;
     timer.callOnTimeout([this](){IndiTimerCallback();});
     timer.setSingleShot(true);
+
+    have_sensor = false;
+    sensor_timer.callOnTimeout([this](){IndiSensorTimerCallback();});
+    sensor_timer.setSingleShot(true);
 }
 IndiAsiPower::~IndiAsiPower()
 {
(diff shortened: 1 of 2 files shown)


tentaisan:
I think the reason why indi-webcam build fails in macos environment is because the API of ffmpeg installed by brew has changed. I made another patch and sent a PR.
#542

ken-self (changes requested):
@tentaisan I had a look at the revised indi-asi-power driver.
1) You should probably use TimerHit for polling the sensor rather than setting up a different timer
2) Use the have_sensor flag in initProperties and updateProperties to conditionally set up the sensor tab
3) You might want to rename one of the n_sensor member properties. I found it confusing when one means "number of sensors" and the other means "sensor number" (suggest sensor_id).

@knro if the sensor tab is made conditional this change looks pretty safe. The i2c bus used for the AsiAir sensor device is bus 10 (x0a) and the standard RPi bus is 0 or 1. I don't have an ASIAir myself so just tested it on a RPi 3 to emulate the older AsiAir Pro which does not have the voltage sensor. When it looks for the sensor on bus 10 it finds nothing and sets have_sensor to false.

knro:
Thanks! Sure, as long it was tested on an actual device and does not interfere with existing device.

tentaisan:
@ken-self Thank you for your review.

> 1. You should probably use TimerHit for polling the sensor rather than setting up a different timer
> 3. You might want to rename one of the n_sensor member properties. I found it confusing when one means "number of sensors" and the other means "sensor number" (suggest sensor_id).

I fixed these.

> 2. Use the have_sensor flag in initProperties and updateProperties to conditionally set up the sensor tab

Currently have_sensor is set using the result of i2c_open() of pigpio. i2c_open() can only be called after pigpio_start () on Connect(). I personally understand that initProperties() is called before Connect(). So I don't think it's possible to behave by referring to have_sensor in initProperties().
I think it is possible to set a new switch on the Main Control tab and use it as a trigger.
Could you explain that a little further?

ken-self:
You are right about pigpio_start. I cannot recall why I put it in Connect rather than in the constructor other than there was an issue when it was in the constructor. You should be able to put the condition in updateProperties which is called on a connect or disconnect. I did a quick trial of that and it worked to remove the tab when there was no sensor.

tentaisan:
I changed updateProperty() to refer to have_sensor and do defineProperty() / deleteProperty(). The sensor tab is displayed depending on the state of have_sensor.

ken-self:
Confirmed that the sensor tab does not show when sensors are not installed. All other functions working correctly.

PR #549, opened 2022-02-22
Astrolink4 : Add 'Auto PWM default ON' settings

blemasle:
Hi,

This PR add a new switch vector `Auto PWM default ON` to the `Settings` tab. When checked, the corresponding `Auto PWM` are enabled on driver start.

I for one systematically forget to start the dew heaters on ekos boot. The `Power default ON` behavior cannot be replicated exactly because it is save in the device, whereas dew heaters configuration is saved in the INDI xml file. Dew heaters therefore start on driver start, not when the device boots. This is really convenient, and close enough to be a worthwhile addition to the driver IMO!

image
image

Same changes as https://github.com/astrojolo/astrolink4usb/pull/4 for astrolink4usb device.

As discussed, I also fixed the driver version declaration which was not using template files

image


knro:
Thanks, did you increase driver version?

blemasle:
> Thanks, did you increase driver version?

🤔 I increased the `VERSION_MINOR` constant but haven't look anywhere else indeed. I'll have a look if anything else needs to be updated as soon as possible.

knro:
Ok that's not reflected in CMakeLists.txt and the driver XML file? usually it's done in the CMakeLists.txt if you check the other drivers.

blemasle:
> Ok that's not reflected in CMakeLists.txt and the driver XML file? usually it's done in the CMakeLists.txt if you check the other drivers.

`setVersion` is called with `VERSION_MAJOR` and `VERSION_MINOR` [here](https://github.com/blemasle/indi-3rdparty/blob/master/indi-astrolink4/indi_astrolink4.cpp#L38).

Are they supposed to match versions declared in the XML and CMakeLists.txt file ? Currently, both are set to `0.1`.
CMakeLists.txt defines constants that are not even use inside the driver itself. 
CMakeLists.txt : [`ASTROLINK4_VERSION_MAJOR`](https://github.com/blemasle/indi-3rdparty/blob/master/indi-astrolink4/CMakeLists.txt#L4)
indi_astrolink4.cpp : [`VERSION_MAJOR`](https://github.com/blemasle/indi-3rdparty/blob/master/indi-astrolink4/indi_astrolink4.cpp#L22)

Looking at the Pegasus Ultimate Power Box driver from which the Astrolink4 driver was I think originally forked it seems that they should match.

I can try to open another PR to sort things out based for instance on the way the [rpi gpio driver does it](https://github.com/indilib/indi-3rdparty/blob/7a6d3dbd86e1d4d99f646c33f0c51078b855c26b/indi-rpi-gpio/CMakeLists.txt), defining constants only once. In this case, i'll just change this PR to make the versions match in all files.

blemasle:
@knro Done.
The driver `.cpp` was not using `config.h.cmake` generated result. Now a simple update inside `CMakeLists.txt` changes all versions usage at once.

Let me know if this should be done differently 🙂

knro:
Usually, you'd have .xml.cmake template file that is updated with the actual version number, check out https://github.com/indilib/indi-3rdparty/blob/master/indi-qhy/indi_qhy.xml.cmake for an example.

blemasle:
Yes that was already the case, but he generated config.h wasn't included so it didn't use the same cmake variable as those in the xml.cmake file.

As  per your example, this is the equivalent on the qhyccd driver https://github.com/indilib/indi-3rdparty/blob/master/indi-qhy/qhy_ccd.cpp#L133

Plus, the generated XML file was in the git index (plus a misnamed XML file as well).

PR #550, opened 2022-02-22
Fix ASICamera version

paolostivanin:
Latest version is 1.21, not 1.21.3 (https://astronomy-imaging-camera.com/software-drivers)


knro:
This would a downgrade, I wonder what will happen upon install? and when packaged in debian packages.

paolostivanin:
Honestly, I don't know, sorry!
I package only stable releases for openSUSE (here the package manager takes care of this situation, so no issues would occur)

PR #556, opened 2022-03-02
Change maximum position for astromechfoc from 9999 to 32767.

gmaxwell:
I was unable to get a 400mm f/2.8 III on a Lens Controller Mark II
to span its entire focus range with the default settings.

With the range [0,32767] it appears to span the whole range, though
it's hard to be sure if the lens is going as far beyond infinity as
lens supports.  Sending 32768 or a higher value to the device makes
it start reporting a negative position number,  so I am pretty sure
that is too far.

### DIFF
--- a/indi-astromechfoc/astromech_focuser.cpp
+++ b/indi-astromechfoc/astromech_focuser.cpp
@@ -37,7 +37,7 @@
 static std::unique_ptr<astromechanics_foc> Astromechanics_foc(new astromechanics_foc());
 
 // Delay for receiving messages
-#define FOC_POSMAX_HARDWARE 9999
+#define FOC_POSMAX_HARDWARE 32767
 #define FOC_POSMIN_HARDWARE 0
 
 bool astromechanics_foc::Disconnect()


gmaxwell:
With some brief googling, I couldn't find any technical documentation that specified
the range.  The protocol docs don't provide one: https://astromechanics.org/downloads/ascom_ef/en/manual/User%20Manual%20PROTOCOL.pdf

So I don't know for sure what the range is supposed to be, I only know that the
current value didn't work for me-- my new value does work for me, and trying to
go any wider appeared to be obviously wrong.

I found another driver at https://github.com/rpineau/EFLensController which also
uses 9999 but it appears to me that in that driver the limit can be toggled off,
so if it's too low it wouldn't be a problem while in the indi driver you can't
go beyond this hardcoded limit not matter what.

Because the device always powers up at a position of 5000 regardless of where the
lens is (the controller doesn't support talking to the encoders that exist on at
least some lenses), the currently limited range may not be fatal for some users--
so long as the start up near infinity and don't need to focus closely, but I can't
think of any reason that losing access to the range would be preferable. Esp. since
you can always reduce it at runtime.

I'll attempt to solicit input from the manufacturer.

gmaxwell:
hold off on this, it still needs work. testing after restarting the controller doesn't get me the same results as controlling it manually via a terminal.  I must have missed something. Will fix.

gmaxwell:
Okay, I don't think the issue is anything in my change--  it's just that the controller comes up with position 5000 from cold start.  If the lens is at infinity, and you go to zero then the move from 5000 to 0 is not enough to make it to zero.  To fix that it's necessary to focus to 0 and 32767 to get the controller centered.   That's unfortunate, but I don't see a way around it: having the driver perform that calibration at startup with no doubt induce homicidal rage in users when it breaks a perfect focus on driver restart. :)  Though maybe I should add an 'initialize' button that seeks to each extreme before returning to whatever value it was at last.

knro:
Thanks! How about contacting Astromech directly?

gmaxwell:
Indeed, I've contacted them and brought this PR to their attention. It would be good to hear from them as I am admittedly clueless-- I just know some fix was necessary for me, and this appears to work. :)

PR #569, opened 2022-04-14
Proposal for automatic temperature compensation on ASI focuser

thx8411:
Hi !

This is a proposal for automatic temperature compensation on the ASI Focuser.

It embeds 3 new settings : 

- steps : steps we should move in or out for each Celsius degree
- hysteresis : minimal temperature change before moving (to avoid yo-yo effect)
- samples : number of samples to average to get stables temperature readings

Uses the local temperature provided by the focuser itself.

Temperature is checked at each "TimerHit" call.

TODO :

- avoid focuser moves during main CCD captures

Best regards !

Thx8411

ekos_temp_comp


knro:
Thanks! I wonder if this is better left off to handle at the client level than at the driver level. Then, the same logic could be used to support any arbitrary stepper focuser. What do you think?

thx8411:
Hi Jasem,

<< What do you think? >>

I don't know, I have mixed feelings, there are pro and cons.
Jarno pointed out some of it in the forum.


Pro :

- Available for all clients
- Some focusers already have built-in temp compensation
- Focuser seen as standalone device

Cons :

- Client side temp compensation enables much richer features (such as "auto learning" facilities, if tied up with the autofocus routine)
- Universal UI, not device dependent
- Avoiding moves while capturing is easy

Thinking long-term, Client Side in Ekos is definitely the best way to do it. But today, the code is available, and might help some of us ;-).

Best regards !

Thx8411

jpaana (approved):
Code looks good to me. I don't think there is any reason why this couldn't exist in both the driver and client and people could use which suits them best. Especially as there isn't anything in Ekos yet :) Having it in the base class might be useful to consolidate the support in various drivers, but that can be next step.

PR #571, opened 2022-04-24
Fixing INDI Webcam Issues

rlancaste:
This should fix several significant issues in INDI Webcam, including
1 - Fixing macOS Monterey crashing issues
2 - Making the IP Camera URL customizable and opening the door to future protocols
3 - Improving reliability and fixing bugs
4 - Making the pixel size for the cameras editable and providing a list to select from
5 - Allowing sub 1s exposures
6 - Improving code to find ffmpeg
7 - Allowing the configuration of the input pixel format
8 - Making it more efficient
9 - Improving messages


knro:
Please check the MacOS failure

rlancaste:
So I don’t know what caused this but I don’t think it is related to my code.

it says 

Error: The log was not found. It may have been deleted based on retention settings.

PR #579, opened 2022-05-20
Enhanced binning for SBIG ST-7/8/etc cameras

tillense:
The SBIG ST-7/8/etc cameras can be used as detectors for the SBIG self-guided spectrograph. Therefore other binning modes with different horizontal and vertical binning are supported by these devices. With this patch all supported binning modes become available.
Currently the code has only been tested with the SBIG ST-8XME. It would be necessary to test it with other SBIG devices before merging with the upstream repository.


knro:
I don't have any SBIG cameras here to test with unfortunately. Code looks good but you can only tell after testing. Let's see if anyone else steps in.

PR #581, opened 2022-05-27
Omegon SDK 51.xxx integration

kneo:
This PR proposes a new binary **indi_omegonprocam_ccd** linking against the new version of the Omegon SDK. 
It also adds support for the Omegon GUIDE 2000 C color guide cam by introducing a new USB Product ID. (There may be more IDs to add, for different versions 1200M/C 2000M).

**Maintainance Work done**:
- Debian Package specs created/updated
- Version Number increased to 0.9 of the toupbase driver
- added binary library for the Omegon SDK
- added PID for the 2000C
- added Omegon Pro/GUIDE entry in the indi xml, so the existing Omegon Pro driver can still be used

**Known issues**:
- when using a Omegon veTEC 571C and the GUIDE 2000C on via hub, connection issues appear. These go away if both cameras are connected to separate ports on the Raspberry PI 4
- In the indi webmanager (with astroberry) restarting the driver states: "error restarting the driver", yet the cam can still be used. This should be looked into, since restarting a hanging camera seems crucial. I don't know whats wrong here, I barely touched the code. 

**Testing done**:
- connected to cameras, and used in parallel
- taken dark library with Omegon Guide 2000C and PHD2
- unfortunatly, the only clearish night in months is accompanied with only high wind gusts, a real life test is not possible for the moment.

clear skies!


bgilsrud:
The changes in the oem_cameras file are not needed. The contents of that file are only useful if you were trying to use an omegon camera (or any other rebranded-touptek) with the official touptek drivers.

knro:
Thank you! looking good.

PR #599, opened 2022-07-03
indi_nut driver (UPS monitor)

rickbassham:
This driver shows up as a "Weather" device and will connect to a NUT ups server to monitor battery status. It goes to unsafe when charge remaining drops below 50%.


rickbassham:
For this driver, since it adds a new library dependency with libnutclient-dev, I thought it was better to include it here. I can still move it if you think it belongs in the main repo.

knro:
Right, the general rule of thumb is that anything with additional dependencies belongs to 3rd party. Let me see.

knro:
Debian files are missing and without them, we cannot add to PPA.

PR #606, opened 2022-07-11
add debian file for indi-weewx-json

rickbassham:


rickbassham:
@knro Is this correct? I haven't dealt with these before.

knro:
Looks good! thanks.

PR #608, opened 2022-07-13
Update QHY SDK to 22.7.6.15

paolostivanin:
SDK download from https://www.qhyccd.com/html/prepub/log_en.html#!log_en.md


knro:
Thank you for the PR, but there is no need, I usually build it from source before each INDI release. Next week I'll update the SDK.

PR #610, opened 2022-07-15
indi-weewx-json: handle large responses properly

rickbassham:
Make sure we are handling large responses from the WeeWX server properly.


knro:
Actually, can you please rework this? I removed gason.h and gason.cpp from original repo and now use a much more modern [JSON parser](https://github.com/nlohmann/json) that's header only. It's installed now by INDI Library so you can make use of it. I've updated the implementation of OpenWeatherMap to make use of this library.

Also working on update Arco rotator to fully use this, so we might as well switch to this method for all JSON parses.

rickbassham:
Will do.

knro:
Thanks! Just checking if there has been any update on this?

rickbassham:
I was waiting of the new json parser to be merged in.

knro:
It's already merged in INDI library, you can utilize it since it's installed to the libindi include directory.

rickbassham:
@knro Updated to use new json parser

PR #615, opened 2022-07-25
ASI Camera SDK 1.25 release

jpaana:
ASI Camera SDK 1.25 release. Hopefully this one fixes the issue with arm64, but not certain in any way.

Changelog including updated entry for 1.24:

V1.25 2022.7.20
Optimize the performance of ASI432

V1.24 2022.6.23
(1) Add ASI585,ASI432,ASI662,ASI678
(2) The camera name from ASIGetCameraProperty no longer contains alias. If the application software needs to display alias, it needs to obtain them through the ASIGetID method.
(3) Some optimizations have been made


knro:
Thanks, I can confirm it's still broken on MacOS. So more than likely broken under arm64 as well.

PR #619, opened 2022-08-01
Add support for C5 cameras, clarify license, extend FITS headers

### LINKED ISSUE #522: indi-mi: missing FITS headers and properties in ekos
**Is your feature request related to a problem? Please describe.**

The Moravian Instruments C4-160000 camera is a dual-gain CMOS camera, which requires taking Dark and Light Frames in two gain settings (read modes). Ideally, I would like to take and preprocess the images with ekos, but currently this is not possible. Instead, SIPS.exe is needed at least to pre-process the images. However, when images are taken with ekos, the sips.exe cannot process because of missing FITS headers indicating the read mode. Likewise, any other program with preprocessing (i.e. PixInsight) will need to know what read mode was used for processing.

As it is right now, the indi-mi driver is useless for the new C4-160000 camera.

**Describe the solution you'd like**
These calibration images are critical for calibration, and we need a few more features to make them work in an automated way:

* please add "Read Mode" from indi_mi Main Control panel to the "Custom Capture Properties" in ekos. This makes it possible to program sequences which automatically take i.e. DARK images in the various Read Modes.
* please add FITS headers for: GAIN and DATAMAX according to this table:

Read Mode | GAIN | DATAMAX
---- | ---- | ----
16-bit HDR | 0.82 | 65535
12-bit hi-gain | 0.82 | 4095
12-bit lo-gain | 18.69 | 4095
"16-bit" lo-gain | 0.82 | 65535

The FITS headers are needed to identify the read mode and also make it possible to pre-process the images with the SIPS.exe from MI.

The FITS header make identification easier and also make output compatible with the MI SIPS.exe program, required for calibration. Perhaps adding a "READMODE" fits header would also help some users.

To fully pre-process the images in ekos, processing must take two DARK and FLAT frames into account. When the measured value (in the 16-bit HDR image) is less or equal to a threshold value, use the 12-bit-hi-gain DARK/FLAT image... Otherwise, use the "16-bit" lo-gain DARK/FLAT images.
The threshold for the C4-160000 camera is 3600 ADU. This is the same procedure taken by the camera, when the HDR mode is used. The documentation for this is in the C4 Camera User's Guide page 43ff:
https://www.gxccd.com/download/C4%20Manual.pdf

**Describe alternatives you've considered**

Currently, my only alternative is to use SIPS.exe instead of Ekos. Or, edit the fits header after recording the files with Ekos.

**Additional context**

I think this new kind of 2-gain cameras will become more common and ekos should be able to pre-process such data as well. Of course, we can make a script to do it outside ekos, but it would be much nicer to have this feature buildin.

sWski:
- updated library adds support for new C5 cameras and optional GPS module attachments
- added license file to clarify distribution
- driver now writes extended info to FITS header, fixes #522


indi-mi/mi_ccd.cpp
```diff
@@ -992,3 +1002,50 @@ bool MICCD::saveConfigItems(FILE *fp)
 
     return true;
 }
+
+void MICCD::addFITSKeywords(fitsfile *fptr, INDI::CCDChip *targetChip)
```
knro, line 1006: Please update the signature of this function to be inline with INDI Library [signature](https://github.com/indilib/indi/blob/328fefd1cfb32a5748a7da1bff441d2b0c3ae2c0/libs/indibase/indiccd.h#L508). targetChip can be accessed like how it is done in [CCD Simulator](https://github.com/indilib/indi/blob/328fefd1cfb32a5748a7da1bff441d2b0c3ae2c0/drivers/ccd/ccd_simulator.cpp#L1494)

sWski: Ah, sorry, didn't have up-to-date indlib sources. It should be ok now.

PR #621, opened 2022-08-03
Improvement: Added driver alias 'SVBONY CCD’.

jctk:
- Added device alias 'SVBONY CCD' to driver name 'SVBONY SV305' in indi_sv305_ccd.xml.cmake.
- Since some users do not realize that the driver name 'SVBONY SV305' is a driver for SV405CC or SV905C, a generic name was set.

### DIFF
--- a/indi-sv305/indi_sv305_ccd.xml.cmake
+++ b/indi-sv305/indi_sv305_ccd.xml.cmake
@@ -5,5 +5,9 @@
                 <driver name="SVBONY SV305">indi_sv305_ccd</driver>
                 <version>@SV305_VERSION_MAJOR@.@SV305_VERSION_MINOR@</version>
 	</device>
+        <device label="SVBONY CCD" mdpd="true" manufacturer="SVBONY">
+                <driver name="SVBONY CCD">indi_sv305_ccd</driver>
+                <version>@SV305_VERSION_MAJOR@.@SV305_VERSION_MINOR@</version>
+	</device>
 </devGroup>
 </driversList>


knro:
Wouldn't adding aliases for  SV405CC and SV905C be more user-friendly? Also these names would be recorded in the FITS header, so useful for later.

jctk:
Registering Aliases for individual product names would be costly in maintenance to add each time new equipment is released in the future.
Since ZWO's driver name is 'ZWO CCD', I used the same method, 'SVBONY CCD'.

The device name set in the driver is not recorded in the FITS Header.
The detected product name is registered in the 'INSTRUME' field of the FITS Header, so it does not affect usability.

knro:
QHY and ZWO drivers rename themselves upon detection the model. Perhaps SVBony driver should do the same if it has the capacity to detect the exact model?

jctk:
When I connect ASI294MC and check with ps command, it is not renamed as indi_asi_ccd.
Also in Profile Editor, ASI294MC is not in the list and remains as ZWO CCD.

INDI_ASI

After the connection, the INDI control panel shows ASI294MC.

This is the same for SVBONY, which becomes the model name in the INDI control panel.

As for the operation, SVBONY and ZWO are the same.

jctk:
This PR is just to make 'SVBONY CCD' appear in the device selection list in the Profile Editor.

Once connected to the camera, the model name is used by the INDI control panel and EKOS.
The model name is also used in the FITS header and config file.

There is no change from the behavior before this PR, except that 'SVBONY CCD' is displayed in the device selection list in the Profile Editor.

thx8411:
Hello Knro and Jctk,

@jctk I also wanted to rename the driver with a more generic name when the SV905C comes out.

Using the SV305 name is confusing for the SV905C and the SV405CC
But using a generic name would be confusing for the SV105 and SV205.

I assumed that SV905C and SV405CC users are more experienced than SV105 and SV205 users, and less confused by the legacy SV305 name ;-)

@knro What do you think ?

Best regards,

Thx8411

knro:
How about just "SVBony"?

jctk:
@knro and @thx8411 
This PR has been rejected ? Is has not been merged.

knro:
Sorry forgot to update this. I made a commit to add 405 and 905 explicitly in this commit https://github.com/indilib/indi-3rdparty/commit/2c8c2fc7f95c8e6a2e2ed1283ca5249bc9849b6b

jctk:
You decided on the policy of adding the actual model name.
It is unfortunate that the generic name was not adopted.

SV505C will be released from SVBONY soon.
I will add SV505C to the alias after the INDI driver is supported.

knro:
The driver needs to be updated:
+ Rename driver to indi-svbony
+ Support multiple cameras by creating unique names like in ASI, QHY, or Touptek camera drivers.

When this is done, we can drop all the aliases and just use SVBony

jctk:
hi, @knro and @thx8411 
I would like to confirm the conditions from knro.

# about driver name change

The options are Plan A, B, or C.
Plan A is the best for the least impact.
Plan C is best in terms of completeness.
Can I proceed with Plan A ? I want to minimize the impact on existing source code.

Plan A

- Make it possible to build with driver name as indi_svbony_ccd.
- Do not change any other folder names/file names/source code.

Plan B

- Allow to build the driver as indi_svbony_ccd.
- Rename the folder from indi-sv305 to indi-svbony by git mv.
- Rename sv305 in the following file name to svbony with git mv.
sv305_ccd.ccp, sv305_ccd.h, indi_sv305_ccd.xml.cmake, indi_sv305.spec
- Allow build to match the above folder/file name changes.
- Do not change the source code such as class name SV305.

Plan C

- Change symbol SV305 to SVBONYCCD in addition to Plan B.

# rename driver to indi-svbony

The behavior is the same as that of the ASI driver if the following behavior is used.
Is this correct?

1. The driver name is indi_svbony_ccd regardless of before or after the camera is connected.
2. The driver choice in the Profile Editor should be "SVBONY CCD".
3. After the connection, the camera name displayed in the INDI control panel and EKOS GUI should be "SVBONY \<model name\> \<connection serial number\>". Example: "SVBONY SV405CC 0".
4. The configuration file shall be "SVBONY \<model name\> \<connection serial number\>_config.xml". Example: "SVBONY SV405CC 0_config.xml".

knro:
Plan C is the correct way to move forward. The only different is that the driver name would be SVBony, no need to add "CCD" to it as this was the legacy way of naming drivers back in the day where CCDs reined supreme.

The rest is absolutely spot on! Thank you for taking the time to work on this, it's greatly appreciated!!

jctk:
I believe that CCD should be added to the class name.
This is to distinguish if SVBONY releases other types of devices in the future (which may or may not be possible).
What do you think?
I am fine either way.

I will postpone this work for now.
I will prioritize the development of SV505C support and support for the new version of SVBONY Camera SDK.

PR #624, opened 2022-08-04
mgen: fix ftdi deprecation.

TallFurryMan:
This PR fixes the use of a deprecated function from the FTDI library v1.5 in the mgen autoguider driver.


TallFurryMan:
@knro it didn't work in CI, somehow. I'll need to try again.

PR #626, opened 2022-08-05
ASI Camera SDK 1.26 release

jpaana:
Looks promising that this might fix the ARM issue!

Changelog:
V1.26 2022.8.4
(1) Fixed the bug of ASI432
(2) Optimized the middle gain of 432;411;461;585;662;678;2600;6200
(3) Fixed the crash problem on ARMV8


jpaana:
According to https://www.indilib.org/forum/general/11931-asi-zwo-driver-keeps-restarting.html?start=48#85070 this seems to fix the ARM64 issue finally!

knro:
Thanks! I'll update the PPA as well.

PR #627, opened 2022-08-05
mgen: fix ftdi deprecation differently.

TallFurryMan:
This PR attempts a different approach to manage FTDI 1.4/1.5 deprecated symbol `ftdi_usb_purge_buffers`.


TallFurryMan:
Of course, it's probably better if I ignore deprecation while defining the weak replacement.

PR #631, opened 2022-08-10
Allow ZWO CCDs to be nicknamed, associated by the serial numbers in their firmware.

murveit:
This will allow multiple cameras of the same type to be used, and have the proper one reliably used, e.g. using the imaging ccd for imaging and the one configured for guiding used for that. Also allows more human-readable device names.

### DIFF
--- a/indi-asi/asi_base.cpp
+++ b/indi-asi/asi_base.cpp
@@ -350,7 +350,7 @@ bool ASIBase::initProperties()
 
     NicknameTP[0].fill("nickname", "nickname", mNickname);
     NicknameTP.fill(getDeviceName(), "NICKNAME", "Nickname", INFO_TAB, IP_RW, 60, IPS_IDLE);
-    
+
     int maxBin = 1;
 
     for (const auto &supportedBin : mCameraInfo.SupportedBins)
@@ -496,8 +496,11 @@ bool ASIBase::updateProperties()
 
         deleteProperty(BlinkNP.getName());
         deleteProperty(SDKVersionSP.getName());
-        deleteProperty(SerialNumberTP.getName());
-        deleteProperty(NicknameTP.getName());
+        if (!mSerialNumber.empty())
+        {
+            deleteProperty(SerialNumberTP.getName());
+            deleteProperty(NicknameTP.getName());
+        }
         deleteProperty(ADCDepthNP.getName());
     }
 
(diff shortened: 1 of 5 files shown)


knro:
@xsnrg Any ideas what's going on in Fedora?

knro (changes requested):

indi-asi/asi_base.cpp
```diff
@@ -464,8 +464,11 @@ bool ASIBase::updateProperties()
         defineProperty(BlinkNP);
         defineProperty(ADCDepthNP);
         defineProperty(SDKVersionSP);
-        defineProperty(SerialNumberTP);
-        defineProperty(NicknameTP);
+        if (!mSerialNumber.empty())
```
knro, line 467: Need to do a similar check in the delete part below.

knro (commented):
Please apply astyle settings to the file.

indi-asi/asi_base.cpp
murveit: Done.

murveit:
Artistic style applied to the 5 files (asi_base.cpp/h, asi_ccd.cpp/h, asi_camera_test.cpp

knro:
@jpaana Any comments from your side? can you please also test to verify?

indi-asi/asi_ccd.cpp
```diff
@@ -129,12 +147,184 @@ static class Loader
...
+                saveNicknames();
+                LOG_INFO("The driver must now be restarted for this change to take effect.");
+            }
+            else
+            {
+                LOG_INFO("Can't apply nickname change--serial number not known.");
+            }
+            NicknameTP.apply();
+        }
+        return true;
```
jpaana, line 300: This return and the next should be inside the brace by one nesting level, now other text edits to the device like observer name are ignored.

jpaana:
Sorry, been busy preparing for family occasions this weekend, but had a short while to test. My cameras are connected to different computers so can't test multiple device case, but otherwise seems to work well apart from the text property issue mentioned above.

indi-asi/asi_ccd.cpp
murveit: @jpaana I moved the return true up one level. I think the one at the bottom is ok where it is. Please double check, and if not, please let me know and I'll change it too.

jpaana: Thanks, works now. I've usually used an early return false at the top of the function if the device is null or doesn't match, as calling the base class function is also unnecessary if the call isn't for our device in the first place, it just does the same check again and returns false.

murveit: I guess I'll leave as is and resolve.

murveit:
@knro Jasem, as I understand it, all issues have been resolved.

PR #637, opened 2022-08-22
Correct miss-model name from SV405C to SV405CC

jctk:


jctk:
Hi, @knro 
I corrected device label for SV405CC.

PR #641, opened 2022-08-23
Update qhy_ccd.cpp

sonny486:
Adding 3x3 Hardware binning mode back.  Is supported in QHY268 and QHY600.

### DIFF
--- a/indi-qhy/qhy_ccd.cpp
+++ b/indi-qhy/qhy_ccd.cpp
@@ -971,11 +971,13 @@ bool QHYCCD::Connect()
 
         ret = IsQHYCCDControlAvailable(m_CameraHandle, CAM_BIN1X1MODE);
         LOGF_DEBUG("Bin 1x1: %s", (ret == QHYCCD_SUCCESS) ? "True" : "False");
-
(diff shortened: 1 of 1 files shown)


knro:
Thanks. This information about which is supported should be then saved and re-used when updateCCDBin is called. If not supported then we return an error. Right now, it just accepts everything even if not really supported.

sonny486:
Yea, I saw it uses softbin if not supported in hardware, however 3x3 was explicitly blocked for some reason.

I am gonna try and clean up some more on the QHY Driver.  I am looking at the Filter being shown as available, however there are no filters, I think this is a simple fix though.

sonny486:
I am gonna clean up some more and see what I can do about the BIN issues.

sonny486:
@knro Is there a mechanism for the indi-driver to pass available binning modes to EKOS?  If so, I can create a way to inquire available binning modes in the driver and pass them to EKOS.  I am not sure EKOS even does this.  I will work on cleaning up the code as best as I can.

knro:
Negative, right now Ekos (and other INDI clients) just check the binning property which is a number property. Another property have to pass this information along somehow.

PR #642, opened 2022-08-24
Update qhy_ccd.cpp

sonny486:
BugFix: Filters shown as active when no filter wheel is present.

Update Binning to include 3x3.  

Bin fail to use software binning.

### DIFF
--- a/indi-qhy/qhy_ccd.cpp
+++ b/indi-qhy/qhy_ccd.cpp
@@ -940,14 +940,19 @@ bool QHYCCD::Connect()
                 {
                     LOG_DEBUG("Camera can support CFW but no filters are present.");
                     m_MaxFilterCount = -1;
+                    HasFilters = false;
                 }
 
                 if (m_MaxFilterCount > 0)
+		{
+                    HasFilters = true;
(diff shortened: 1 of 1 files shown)


knro:
No need to inquire for binning support twice,we can do it once. Create a a variable 

bool m_SupportedBins[4];

m_SupportedBins[] = IsQHYCCDControlAvailable(m_CameraHandle, CAM_BIN2X2MODE)

sonny486:
> No need to inquire for binning support twice,we can do it once. Create a a variable
> 
> bool m_SupportedBins[4]; enum { Bin1x1, Bin2x2, Bin3x3, Bin4x4, }
> 
> m_SupportedBins[Bin2x2] = IsQHYCCDControlAvailable(m_CameraHandle, CAM_BIN2X2MODE);
> 
> Then you can use this to see whether there is native support or fall back to software. Did you test 3x3 in software, does it work?

I will have to create a test run on software binning test due to all my QHY Cameras have bin1 - 4.  I will double check.

I will add some logging to test software binning.

sonny486:
Working on the new one now

PR #643, opened 2022-08-24
Update qhy_ccd.cpp

sonny486:
@knro I don't know why, but the write functions of the resolutions are not functioning now, I have no idea why though.  Will work some more on this.  

I do apologize, I am not a programmer by trade, but I am learning and I have the SDK documentation, I will do my best.  I haven't written a c program since Borland C was around.


sonny486:
@knro Also I did check the BIN modes using my QHY163 - it only does BIN 1X1 - BIN2X2.
Binning in 3x3 or 4x4 throws an error.  I don't know why softbin isn't working.

PR #651, opened 2022-09-09
BUGFIX: change mutex life cycle to camera connection period and correct some pthread function calls

jctk:
BUGFIX: mutex life cycle to camera connection period
- Once the camera is closed and reconnected to the camera, the driver freezes without unlocking the mutex for the next camera disconnect.
- Corrected the life cycle of mutex, cond and thread of pthread from the scope of existence of Sv305CCD object to the scope of Sv305CCD::Connect to Sv305CCD::Disconnect.
- Before the correction, the state of the previous mutex was inherited at the second connection, causing a freeze on pthead_lock at the time of Disconnect.

BUGFIX: cancel streaming thread at Disconnect camera.

BUGFIX: some pthread_cond_signal was moved within the period when mutex is locked.
- pthread_cond_signal must be called during mutex lock.

BUGFIX: Fixed a case where a function exits within the mutex lock period of streamVideo.

### DIFF
--- a/indi-sv305/sv305_ccd.cpp
+++ b/indi-sv305/sv305_ccd.cpp
@@ -87,19 +87,12 @@ Sv305CCD::Sv305CCD(int numCamera)
     // Set camera name
     snprintf(this->name, 32, "%s %d", cameraInfo.FriendlyName, numCamera);
     setDeviceName(this->name);
-
-    // mutex init
-    pthread_mutex_init(&cameraID_mutex, NULL);
-    pthread_mutex_init(&streaming_mutex, NULL);
 }
 
 
 //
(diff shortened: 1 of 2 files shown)


jctk:
Hi, @knro
I did another pull request.
Could you please review it?

This time it's a fix for a bug related to mutex.

Fixed a freeze when disconnecting from the camera once without exiting the driver and then disconnecting from the camera a second time.
This was caused by mutex not being reinitialized when the camera was connected.

Other pthread and mutex related bugs have been fixed.

Best regards
jctk

knro:
Thanks I can't test now but it looks good to go, thanks!

jctk:
Thanks for the review and merge!

PR #656, opened 2022-09-17
Update dreamfocuser.cpp

pidlug:
Change FOCUS_WEATHER and related properties names to allow Ekos to find the temperature.


knro:
WEATHER_PARAMETERS property is used for weather device. For focusers, it's expected in FOCUSER_TEMPERATURE. So perhaps just create FOCUSER_TEMPERATURE number property with one element to hold the temperature. Then the rest can remain in the existing property.

pidlug:
Ok. I'll fix that. Thanks!

Issue #654, opened 2022-09-17
indi_sv305_ccd: Last line has garbage data


jctk:
**Describe the bug**
When Frame mode is set to RAW8 or Y8 in indi_sv305_ccd, the last line of data after shooting is garbage data.

**To Reproduce**
1. Run driver
2. Set the Frame mode to RAW8 or Y8.
3. When taken, the last line of data is random patterns.

**Expected behavior**
The last line contains the image data taken.

**Screenshots**
Arrows point to destination line:
![image](https://user-images.githubusercontent.com/52247294/190856600-0d05987d-9442-421f-9ae1-c94b448a351f.png)

**Desktop (please complete the following information):**
 - INDI on OS: StellarMate OS
 - Client KStars on OS: Windows 11
 - INDI Version 1.9.7

**Log Files**
[log_21-29-12.txt](https://github.com/indilib/indi-3rdparty/files/9591664/log_21-29-12.txt)

jctk:
Hi, @knro and @thx8411 
I will investigate when I have time.

jctk:
Additional evidence.
Reproduces even in Frame Format RAW16
![image](https://user-images.githubusercontent.com/52247294/190904854-65383244-176b-4498-a1e3-556a7d73f8f2.png)

jctk:
## Problem Cause ##
After setting the ROI with SVBSetROIFormat, the actual ROI must be obtained with SVBGetROIFormat.
The ROI obtained by SVBGetROIFormat has less height than that specified by SVBSetROIFormat.

In sv305_ccd.cpp, the process to get the ROI actually set by SVBGetROIFormat is not implemented.

## See also ##
SVBDemo.cpp's part on setting ROI.

PR #658, opened 2022-09-19
BUGFIX: indi_sv305_ccd: garbage data in last line (Issue #654)

jctk:
- In sv305_ccd.cpp, I implemented a process to get the ROI actually set by SVBGetROIFormat.
- After setting the ROI with SVBSetROIFormat, I need to get the actual ROI with SVBGetROIFormat, because the ROI obtained with SVBGetROIFormat may differ from the value specified with SVBSetROIFormat.
- Issue #654 was raised because this feature was not implemented.

### DIFF
--- a/indi-sv305/sv305_ccd.cpp
+++ b/indi-sv305/sv305_ccd.cpp
@@ -592,16 +592,27 @@ bool Sv305CCD::Connect()
 
     // set camera ROI and BIN
     binning = false;
-    SetCCDParams(cameraProperty.MaxWidth, cameraProperty.MaxHeight, bitDepth, pixelSize, pixelSize);
     status = SVBSetROIFormat(cameraID, 0, 0, cameraProperty.MaxWidth, cameraProperty.MaxHeight, 1);
     if(status != SVB_SUCCESS)
     {
-        LOG_ERROR("Error, camera set ROI failed\n");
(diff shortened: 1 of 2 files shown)


jctk:
Hi, @knro 
I would like to know how to set values for Frame X: Y: Size Width: Height: in Ekos. This is to reflect the correct ROI.

Top Left Width Height in INDI Control Panel can be set by member functions of CCD or by `CCD_FRAME` in Standard Properties.
These methods cannot be reflected in Ekos.

image

By the way, when I changed the Height in the INDI Control Panel on the ASI CCD, it was not reflected in Ekos.
Is there any way to set this in Ekos?

knro:
Thanks, this is not an INDI issue, but it appears you're using a slightly older version of KStars? have you tried 3.6.0 stable or beta 3.6.1 ?

jctk:
I am using KStars 3.6.0 Stable.
I have tested KStars on Windows 11 and StellarMate OS.

jctk:
The screenshots are transcribed from the kStars website.

knro:
Ah okay, the screenshots are out of date. At any rate, the ROI in Ekos reflects *target* setting and not necessary current camera settings.

jctk:
Thanks for the review and merge.
I understand that the ROI for Ekos is target setting.

PR #661, opened 2022-09-21
Restrict overmatching MACH ifdef to only trigger on OSX and Mach

Zopolis4:
As said on lines 4 and 6, this check is only intended for OSX.
```
/* Mac OSX prior Sierra is missing clock_gettime() */
```

Hurd also uses Mach, but does not have the same shortcomings as OSX in this area.
https://www.gnu.org/software/hurd/hurd/porting/guidelines.html
```
#ifdef __MACH__
Some applications put Apple Darwin-specific code inside #ifdef __MACH__ guards. Such guard is clearly not enough, since not only Apple uses Mach as a kernel. This should be replaced by #if defined(__MACH__) && defined(__APPLE__)
```

### DIFF
--- a/indi-eqmod/mach_gettime.cpp
+++ b/indi-eqmod/mach_gettime.cpp
@@ -1,7 +1,7 @@
 /* complete rewrite May 2017, Rumen G.Bogdanovski */
 #include <time.h>
 
-#ifdef __MACH__ /* Mac OSX prior Sierra is missing clock_gettime() */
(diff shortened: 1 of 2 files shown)


knro:
@xsnrg Can you please check the artifacts errors? This is new.

xsnrg:
@knro I do not think this is new. I have not seen Fedora artifacts get created on any of my PRs either, and I did not change anything that would have affected artifact creation in those PRs. It looks like there may be more setup needed for the Fedora part on github.

This PR looks okay to me. The Fedora build did complete here on github as well, just no artifacts were created.

PR #672, opened 2022-10-15
Svbony driver name generalization

jctk:
- The driver name for SVBONY is indi_sv305_ccd. Change it to indi_svbony_ccd.
- The class name, file name, directory name, and various other symbols are subject to change.


jctk:
Hi @knro 

I generalized the SVBONY CCD driver name from the model name.

But Pull Request Check failed to build.
I git mv'd indi-sv305 and libsv305 to indi-svbony and libsvbony.
Maybe indi-sv305 and libsv305 are still in the build check environment?
Can you please confirm that the git mv deletion of the previous directories is reflected in the pull request?

debian/libsvbony/changelog
```diff
@@ -1,3 +1,23 @@
+libsv305 (1.9.6) bionic; urgency=low
```
knro, line 1: This should be libsvbony

knro (changes requested):
Looks good! Perhaps a small suggestion, maybe camera names should start with index 1? i.e. SVBony SV305 1 instead of zero. 0 makes sense for programmers, but for users, starting with 1 makes more sense. Small change requested in changelog.

debian/libsvbony/changelog
jctk: Thanks for the review.
Changelog has been corrected.

Can you make the Camera Number a Feature request as an Issue? I would like to consider it separately from this Pull Request.

By the way, if this PR is merged, what will happen to indi_sv305_ccd already installed in the user's environment ? Will it be automatically removed by apt upgrade ?

knro:
No it will not be removed. You can add a postint script to Debian directory to remove /usr/bin/indi_svbony_ccd if found.

jctk:
I added `postinst` to the directory `debian/ind-svbony` as follows.
Is this ok?

```sh
#!/bin/sh
# postinst script for indi_svbony_ccd

if [ -f /usr/bin/indi_sv305_ccd ]; then
    rm -f "/usr/bin/indi_sv305_ccd" 2> /dev/null
fi
```

knro:
Yes this looks good!

jctk:
I forgot that the following files also need to be deleted in postinst.
`/usr/share/indi/indi_sv305_ccd.xml`

jctk:
Hi @knro 

Can you please modify the following content to reflect the driver name change?

https://www.indilib.org/ccds/svbony-sv305-cameras.html

Create svbony-ccd-cameras.html.
And svbony-sv305-cameras.html should be forwarded to svbony-ccd-cameras.html.

knro:
Sure it's done.

jctk:
It's true, it's updated. Excellent!
When did you...

If you don't mind, could you replace the image as well?

knro:
Replace with what? please put one.

jctk:
This
https://github.com/indilib/indi-3rdparty/blob/master/indi-svbony/doc/control_panel.jpg
- frame formats has been moved to option panel.


and this
https://github.com/indilib/indi-3rdparty/blob/master/indi-svbony/doc/info_panel.jpg
- driver name has been changed from indi_sv305_ccd to indi_svbony_ccd

PR #679, opened 2022-10-30
Add Excalibur Driver

rbfocus:
Adding RB Focus Excalibur driver to 3rd party Repo


knro:
Thank you! If this driver does not have any external dependencies then it should submitted to core INDI library drivers. Maintaining 3rd party drivers is a lot of work and reserved for drivers with external dependencies not met in INDI core.

rbfocus:
ok I will configure everything for INDI core. Thank you.

PR #697, opened 2022-12-21
Libcamera

anjok:


knro:
Thanks! why is libcamera-apps included in the source? I think it should be used as an external dependency since it is a separate and fast moving project as well.

anjok:
It IS apparently a fast moving project, and I for my part would prefer a static end result instead of dealing with compile-from-source lib versions. When they stabilise or actually do anything new we actually need, we can use the libs again.

Also, they didn't expose the CameraManager, which is needed to get at the cam list. It's supposed to be a singleton and this was leading to crashes. Sure I could have posted a PR there and let them handle it, but for now my focus has been on getting things to work in the first place.

And finally this thing crashed left and right for all sorts of reasons and it was my hope to get at least some context info by having the source included.

A probably better way would be to strip out the parts we need and do things directly, not using libcamera-apps, but that again would be a "copy", albeit a possibly buggy one. 

So copying the source was the most economic choice IMO.

anjok:
If you want you can start the workflows, but don't merge just yet. There's issues with the framerates I hopefully get to work today or tomorrow.

knro:
@anjok How is it going? I see a lot of progress in the INDI forum thread.

knro:
@anjok is this driver *usable* at this stage?

anjok:
@knro it didn't crash in a while, but I didn't get around to do as much testing as I'd wish. And it's still weird in that raw pics don't take up changed settings unless you change them in jpg mode first. Also, since it includes the libcamera apps code, you need to pin it to specific versions of boost and whatever else it needs. But I guess it'd be the same with the shared apps libs.

It takes take pics, most of the settings work and you can "use" it though.

knro:
Ok, perhaps we can merge it now so we can get more feedback from users?

anjok:
In your place, I'd probably build and test it locally, but sure, go ahead.

PR #700, opened 2023-01-08
Smooth warming for ASI CCD

pmneo:
Because of the silicon oil issue of the asi 2600 cameras, I have implemented a smooth warming procedure, if cooler is switched off. 

In this case the temperature is stepped in 2 degree up (has to settle for at least aprox. 5 seconds) until the temp does not reach the next 2 deg step within 5 minutes. This indicates the end of the warming procedure.


knro:
I believe this should be in INDI::CCD layer and not just limited to ZWO. There is already a gradual cooling functionality in the CCD (degrees per minute). Maybe on setting a high temperature, it should gradually warm until reaching the ambient temperature?

pmneo:
Yes, I have seen that function of ramping too. But to implement it clean like this, the indicdd should define the enableCooler switches and delegation to enableCooler like SetTemperature ...

knro:
Any chance you can re-work this for INDI::CCD? We can't just have it for this camera.

luzpaz:
bumping

PR #702, opened 2023-01-19
Add PlayerOne filter wheel driver and Update camera SDK to v3.1.1

hiro3110i:
PlayerOne Phenix filter wheel driver is added.
And camera SDK is updated to v3.1.1 which is fixed Sedna-M issue.
https://github.com/indilib/indi-3rdparty/issues/684


knro:
Thank you!

PR #707, opened 2023-01-22
Binning Mode Feedback

42-max:
User Denis from the INDI forum found that there is no feedback from the binning mode setting in the Touptek Indi driver.


knro:
Can you elaborate? I don't understand the problem.

PR #710, opened 2023-01-25
Touptek Driver: Allow Window Heater Levels

42-max:
In the INDI forum user Lanco wrote that some Toupcams allow different heater levels. The pull request implements this heater levels.

Screeshot_HeatLevel

The Heater on/off button will now switch the heater to the selected level.
 
Level can be adjusted while heater is on or off.

Testing was done by measuring the power supply current for different levels.

### DIFF
--- a/indi-toupbase/indi_toupbase.cpp
+++ b/indi-toupbase/indi_toupbase.cpp
@@ -152,6 +152,7 @@ bool ToupBase::initProperties()
 {
     INDI::CCD::initProperties();
 
+    LOG_INFO("init properties");
     ///////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


knro:
FOr new properties, please use new INDI properties. i.e. 

INDI::PropertySwitch

Also, it is customary to have this switch as Enabled/Disabled or On/Off and not the other way around.

42-max:
The idea was to change as less as possible... I will try to use the new property. Are you OK with buttons to switch on/off and an additional level property? The level alone [0..max] would also work as 0 means heater off. But this would cause changes on client side.

knro:
Hopefully @touptek can implement this feature now that they maintain the driver.

PR #730, opened 2023-02-15
Touptek step 3

touptek:


knro:
Great job! However, I don't think  8da2c5b is OK. You should let INDI control this as per the user. This feature is disabled by default. If the user enables it, then the driver should just cool to the temperature as sent by the INDI::Camera class as is. This is how all other drivers behave.

touptek:
> Great job! However, I don't think [8da2c5b](https://github.com/indilib/indi-3rdparty/commit/8da2c5b59a5d6d4dd17e3929c296282bd4618fab) is OK. You should let INDI control this as per the user. This feature is disabled by default. If the user enables it, then the driver should just cool to the temperature as sent by the INDI::Camera class as is. This is how all other drivers behave.

ok. I'll revert it

touptek:
On a x64 computer, stream more than 5 hours continuously using RGB format, there is no crash.
Can you have a try? @knro

touptek:
Have another about 3 hours test, still ok. @knro

knro:
Thanks, I will check this soon!

PR #733, opened 2023-02-17
Pass timestamp value into SER recorder

nouspiro:
Pass exposure start timestamp from GPS into SER timestamp.


nouspiro:
I tested this with one owner of QHY174GPS and one thing we noticed is that start and end timestamps are within one millisecond from each other.

knro:
Great, but I was wondering if anyone with GPS would benefit from this instead of just making it specific to QHY?

nouspiro:
I also changed that latitude and longitude is stored in FITS as decimal degrees.

PR #746, opened 2023-03-01
Touptek bugfix: timer start use milliseconds

touptek:


sonny486:
@touptek @knro   is this good to commit yet?

I will test on my setup.

knro:
Go ahead test and report back

PR #752, opened 2023-03-10
Send AUX commands to MC imitating attached HC

tstibor:
When no HC is attached, the following three commands needs to be send to the motor controller (MC): MC_SET_POSITION, MC_SET_CORDWRAP_POSITION and MC_CORDWRAP_ON. These three commands are also send by the HC to the MC during HC startup and quick align process.

### DIFF
--- a/indi-celestronaux/celestronaux.cpp
+++ b/indi-celestronaux/celestronaux.cpp
@@ -134,7 +134,7 @@ bool CelestronAUX::Handshake()
 
                 LOG_INFO("Setting serial speed to 9600 baud.");
 
-                // detect if connectd to HC port or to mount USB port
+                // detect if connected to HC port or to mount USB port
                 // ask for HC version
                 char version[10];
                 if ((m_isHandController = detectHC(version, 10)))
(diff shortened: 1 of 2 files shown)


indi-celestronaux/celestronaux.cpp
```diff
@@ -513,6 +513,17 @@ bool CelestronAUX::updateProperties()
...
+        // When no HC is attached, the following three commands needs to be send
+	// to the motor controller (MC): MC_SET_POSITION, MC_SET_CORDWRAP_POSITION
+	// and MC_CORDWRAP_ON. These three commands are also send by the HC
+	// to the MC during HC startup and quick align process.
+        // TODO: One can set the HC in pass through mode, that is,
+        // the HC relays the AUX commands only and does not interfere in the communication.
+	if (!m_isHandController && m_MountType == EQ_GEM)
+	{
+	    startupWithoutHC();
+	}
```
knro, line 525: Why limit to EQ_GEM, we can make it work for all types perhaps in startupWithoutHC can have this?

data[0] = (m_MountType == EQ_GEM) ? 0x40 : 00

wouldn't that be sufficient?

tstibor:
I will fix this in the next commit. Yes, I forgot about the 0°/0° for non-GEM mount.

knro:
nevermind, I will push a fix in like a minute

tstibor:
Ok perfect !!!

PR #753, opened 2023-03-11
indi-rpicam: Set the JPEG quality to 1 because it will be skipped anyway

IanCassTwo:
More progress in my effort to improve performance of this driver. Set the JPEG quality to 1 because we're skipping the JPEG data anyway before we extract the RAW data. This results in ~220k less data on a 2028x1520 image and is ~0.02secs quicker to process.


knro:
Thanks, are there any major issues with the current driver now? I wasn't able to test it.

IanCassTwo:
You mean in general or related to my PR's? On the whole, the driver works well except for that irritating issue where long exposure time doesn't work unless your first exposure is at least 7 seconds. Performance wasn't great hence these PR's. Previously it would take ~3-4 seconds to obtain a 1 second exposure which wasn't fantastic especially if you were using it as a guide cam. My PR's have improve that a lot. The thing I'd like to work on next is getting 2x2 binning to work from the UI without hard coding it.

PR #765, opened 2023-03-22
ZWO ASI EAF SDK v1.6

const86:
EAF SDK v1.6 can work with new EAF firmware v3.3.7 to support
focuser range larger than previously used 60000.

Validated with my EAF 5V upgraded to 3.3.7. Set max steps to 200000.
Now it can cover the entire Edge HD focus range in one go.


jpaana (changes requested):
Seems to be missing x64 library. Also updating the debian files would be nice, though I couldn't find any change log from ZWO site.

const86:
Oops, added the x64 binary too.
Right, I didn't find any official change log for both the SDK and the firmware. What exactly should I change in debian files?

PR #776, opened 2023-04-19
Add rolloffino to 3rdparty drivers

wotalota:


wotalota:
A Dome based roll off driver that uses an Arduino to control the roof.

knro:
Thank you! So the arduino controls a relay to open/close and a sensor to know if fully closed or fully opened?

wotalota:
Yes that is the most common approach. Giles modified the code to use linear actuators. There is also an example of using a small DC motor by changing voltage and polarity. The driver just sends open or close and requests state corresponding switch/sensor.

PR #783, opened 2023-05-09
Astroasis INID Driver

astroasis:
This implements Astroasis INDI driver for Oasis Focuser


knro:
Thanks, we also need to add documentation for this to the INDI Website.

PR #792, opened 2023-05-31
Sanitize response eqmod/skywatcher response data

praetp:
I noticed many responses from the mount contained none-hexadecimal characters pointing to some kind of corruption. By rejecting such responses and forcing the client to retry, no corrupt data will propagate to the higher layers.

This should improve situation as described in https://www.cloudynights.com/topic/749844-eq6-r-pro-tracking-glitch/ This will allow people with buggy decoders to use safeguards such as HA limits again.

### DIFF
--- a/indi-eqmod/skywatcher.cpp
+++ b/indi-eqmod/skywatcher.cpp
@@ -2011,8 +2011,9 @@ bool Skywatcher::dispatch_command(SkywatcherCommand cmd, SkywatcherAxis axis, ch
                 return true;
             }
         }
-        catch (EQModError)
+        catch (EQModError ex)
         {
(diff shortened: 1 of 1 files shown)


knro:
Thanks! Need to let this run for a few days in my home observatory with EQ8. I presume you tested this already?

praetp:
I have tested this with my HEQ5-R pro and it makes a big difference.

However, I do not think this will solve all problems: if the corruption happens to generate another "hex" character, it will not be detected.

I think we require a second fix (for RA and DE encoder) where we detect if the jump in retrieved RA/Dec values are not realistic based on current rate.

praetp:
@knro Did you manage to test this ?

knro:
Sorry just tested now and everything appears in order! Let's see if we get any user reports

PR #809, opened 2023-07-02
Update 99-ogmacam.rules

sonny486:
Add TP AP26MC Camera USB ID to rules.  

My permissions were not correct with the Ogma IMX571 Mono camera I just got.


knro:
Thanks for the update!

PR #810, opened 2023-07-02
SVBONY CCD firmware and SDK version information logs added

jctk:
Added code to SVBONYCCD::Connect() to output the version of SVBONY Camera's firmware and SDK as information log.
This information is used when users inquire about issues.

### DIFF
--- a/indi-svbony/svbony_ccd.cpp
+++ b/indi-svbony/svbony_ccd.cpp
@@ -226,7 +226,7 @@ bool SVBONYCCD::Connect()
         LOGF_INFO("Camera Firmware Version:%s", cameraFirmwareVersion);
     }
     else {
-        LOG_ERROR("Error, getting Camera Firmware Version failed.\n");
+        LOG_ERROR("Error, getting Camera Firmware Version failed.");
     }
     // Get SVBONY Camera SDK Version
     SDKVersion = SVBGetSDKVersion();
--- a/indi-svbony/svbony_ccd.cpp
+++ b/indi-svbony/svbony_ccd.cpp
@@ -219,6 +219,19 @@ bool SVBONYCCD::Connect()
         return false;
     }
 
+    // Get Camera Firmware Version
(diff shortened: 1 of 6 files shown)


indi-svbony/svbony_ccd.cpp
```diff
@@ -219,6 +219,19 @@ bool SVBONYCCD::Connect()
...
     }
 
+    // Get Camera Firmware Version
+    status = SVBGetCameraFirmwareVersion(cameraID, cameraFirmwareVersion);
+    if (status == SVB_SUCCESS)
+    {
+        LOGF_INFO("Camera Firmware Version:%s", cameraFirmwareVersion);
+    }
+    else {
+        LOG_ERROR("Error, getting Camera Firmware Version failed.\n");
```
knro, line 229: No need for \n

indi-svbony/svbony_ccd.h
```diff
@@ -109,6 +109,11 @@ class SVBONYCCD : public INDI::CCD
         // pixel size
         float pixelSize;
 
+        // Camera Firmware version number
+        char cameraFirmwareVersion[65];
+        // SVBONY Camera SDK version number
+        const char* SDKVersion;
```
knro, line 115: If you're only using this to report version information once in Connect, is there a reason to keep it as class variable? or is this used elsewhere?

indi-svbony/svbony_ccd.cpp
jctk: I committed it with newline code removed form LOG_ERROR.

indi-svbony/svbony_ccd.h
jctk: If EKOS is used, this information is output to the log and need not be retained.
If non-EKOS clients use it without logging, I will retain this information for reference when debugging.
We also plan to display this information in the INDI control panel in the future, so we decided to keep it.

PR #815, opened 2023-07-19
Add sensor mode control for Player One cameras

jpaana:
After 6 years with Atik 383L+ I finally upgraded my imaging camera to Player One Poseidon-M PRO. While playing around with it I noticed that setting sensor mode of the camera was not supported, which was understandable as most cameras don't have more than one, but Poseidon has two modes, low noise and normal. The SDK has nice functions to enumerate the modes and even get some description for them so adding the functionality was very straight forward.

Additionally I made some clean ups and ran through astyle which caused some diff noise to old code.


knro:
Great work @jpaana

PR #826, opened 2023-08-25
update Touptek & oem lib, version: 54.23231.20230823

touptek:


sonny486:
@knro Here is the Toupbase fix.  thanks.

PR #842, opened 2023-09-26
Add Oasis Filter Wheel driver and update compilation and installation instructions for indi-astroasis

astroasis:
Update for compilation and installation instructions for indi-astroasis


knro:
Thank you. Please update the debian files as well.

astroasis:
> Thank you. Please update the debian files as well.
Thanks for your reminder! As I am not familiar with Debian, could you please let me know which particular files I need to update? And if there is any common rules that I can follow to update those files?

Also could you please merge the current PR first and I will update debian files and create another PR later. This is because I have committed more changes on my forked indi-3rdparty source tree. After the current PR is mergered I will create another PR for those changes and then update debian files. I think this can avoid the risk of messing up the code and PRs.

astroasis:
> Thank you. Please update the debian files as well.

Thanks for your reminder! As I am not familiar with Debian, could you please let me know which particular files I need to update? And if there is any common rules that I can follow to update those files?

Also could you please merge the current PR first and I will update debian files and create another PR later. This is because I have committed more changes on my forked indi-3rdparty source tree. After the current PR is mergered I will create another PR for those changes and then update debian files. I think this can avoid the risk of messing up the code and PRs.

knro:
Sure, all  you need to do is update the [Changelog](https://github.com/indilib/indi-3rdparty/blob/master/debian/indi-astroasis/changelog) file to reflect the updated version.

astroasis:
> Sure, all you need to do is update the [Changelog](https://github.com/indilib/indi-3rdparty/blob/master/debian/indi-astroasis/changelog) file to reflect the updated version. You do this by adding an entry above the current one and just change version and date/time in the same format.

Thanks. Now I have updated and commited the changelog file. Also I have changed the title of this PR from
"Update for compilation and installation instructions for indi-astroasis"
to
"Add Oasis Filter Wheel driver and update compilation and installation instructions for indi-astroasis"

knro:
Thank you. debian/libastroasis probably needs to be updated as well since you updated the library.

astroasis:
> Thank you. debian/libastroasis probably needs to be updated as well since you updated the library.

Now I have updated and committed debian/libastroasis/changelog. Thanks!

PR #845, opened 2023-10-01
Stabilization of SVBONYBase::workerExposure

jctk:
svbony_base.cpp, svbony_base.h
- SVBONYBase::grabImage disassembled and embedded in SVBONYBase::workerExposure
- Remove unused SVBONYBase::grabImage function

svbony_helpers.h
- Added processing in preparation for monochrome camera support (monochrome support not yet completed)
- Changed Luma to Luma 8 bit and Luma 16 bit

### DIFF
--- a/indi-svbony/svbony_base.cpp
+++ b/indi-svbony/svbony_base.cpp
@@ -125,7 +125,7 @@ void SVBONYBase::workerExposure(const std::atomic_bool &isAboutToQuit, float dur
         if (ret == SVB_SUCCESS)
             break;
 
-        LOGF_ERROR("Failed to start exposure (%d)", Helpers::toString(ret));
+        LOGF_ERROR("Failed to start exposure (%s)", Helpers::toString(ret));
         // Wait 100ms before trying again
         usleep(100 * 1000);
     }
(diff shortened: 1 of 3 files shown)


knro:
This was tested on which camera models? did it include video streaming as well? aborted capture? changing formats?

jctk:
I did the test excluding Video Stream. I tried it a little.
* I have not edited the code for Video Stream before or since.

**Testing information:**

Camera model:
- SV405CC
- SV905C

Test Variations:
- Capture a preview and Abort capture.
- Start Framing (Looping)
- Sequence (Raw 8 x1 -> Raw 16 -> Luma 8 -> RGB 24
- Simultaneous exposure. "Start Framing (looping) by SV405CC" and "Guide loop by SV905C"

Video Streaming
- The color image is not correct before the still image is taken.
- I assume that this is because the code after refactoring does not determine the Bayer Pattern until a still image is taken.
I want you to Fix this. I think it is due to the fact that I asked you to fill in the Bayer Pattern in advance but you decided not to do so.

knro:
For streaming, if you capture an image first then stream, the colors are OK?

jctk:
It looked that way to me.
I streamed the entire iPad screen in red and held it up to a telescope.
Before the still image was taken it was green and after it was taken it was red.

jctk:
Could you please take care of Video Streaming after you handle this PR?

jctk:
Thanks for merging the PR.

PR #846, opened 2023-10-08
Update for coding style

astroasis:
Update for coding style


knro:
You're using astyle for format it?

astroasis:
> You're using astyle for format it?

Yes, I was using astyle for the coding style with the ~/.astylerc file recommended on page https://github.com/indilib/indi-3rdparty
Please let me know if this is the correct way for the coding style.

But this is the first time that I use astyle for the commits. I think I should use astyle from the beginning.

PR #850, opened 2023-10-15
update Touptek & oem lib, version: 54.23585.20231015

touptek:


sonny486:
@touptek Does this update include the Sony 585 chip cameras?

sonny486:
@knro This is good to commit.

PR #851, opened 2023-10-22
Modified to be buildable for libcamera 0.1.0

pumomochin:
indi-libcamera: As for removed alias of StreamRoles in libcamera 0.1.0, these code are compiled well for this environment, which is on raspberry pi 4  with imx477.


knro:
Any chance we can completely remove the embedded libcamera and rely on external library?

pumomochin:
Thank you for your reply. The pointed-out embedded libcamera is not found in this project; this project may depend on the external library, which is libcamera. I didn't build for another version of libcamera. However, my pull request for applying to the libcamera update is reasonable because this external libcamera update only removed the alias of the definition.

PR #854, opened 2023-11-17
Adding AvalonUD support

AvalonService:
This is the initial public release of the AvalonUD (Avalon Unified Drivers) drivers that are compatible with the Avalon Instruments StarGO2Pro and StarGO+ controllers.

The release version number is 2.2 since it is aligned to the controllers internal software/firmware version.

It requires the ZeroMQ networking library and for this reason we tried to include it in the requisites script and added a new cmake_modules module.

A CMakeLists.txt file update has also been proposed to enable driver compilation.

We succefully tested build on 32bit Raspbian, 64bit Ubuntu and 64bit Fedora.

### DIFF
--- a/cmake_modules/FindZMQ.cmake
+++ b/cmake_modules/FindZMQ.cmake
@@ -1,0 +1,49 @@
+# - Try to find ZMQ
+# Once done this will define
+#
+#  ZMQ_FOUND - system has ZMQ
+#  ZMQ_INCLUDE_DIR - the ZMQ include directory
+#  ZMQ_LIBRARIES - Link these to use ZMQ
+
+if (ZMQ_INCLUDE_DIR AND ZMQ_LIBRARIES)
+
+  # in cache already
+  set(ZMQ_FOUND TRUE)
+  message(STATUS "Found libzmq: ${ZMQ_LIBRARIES}")
+
+else (ZMQ_INCLUDE_DIR AND ZMQ_LIBRARIES)
+
+  find_path(ZMQ_INCLUDE_DIR
+    NAMES zmq.h
+    ${_obIncDir}
+    ${GNUWIN32_DIR}/include
+  )
+
+  find_library(ZMQ_LIBRARIES
+    NAMES libzmq3
+    PATHS
+    ${_obLinkDir}
+    ${GNUWIN32_DIR}/lib
+  )
+
+
+  if(ZMQ_INCLUDE_DIR AND ZMQ_LIBRARIES)
+    set(ZMQ_FOUND TRUE)
+  else(ZMQ_INCLUDE_DIR AND ZMQ_LIBRARIES)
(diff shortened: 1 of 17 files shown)


indi-avalonud/indi_avalonud_aux.h
```diff
@@ -0,0 +1,126 @@
...
+    ISwitchVectorProperty OUTPort2SP;
+
+    ISwitch USBPort1S[2];
+    ISwitchVectorProperty USBPort1SP;
+
+    ISwitch USBPort2S[2];
+    ISwitchVectorProperty USBPort2SP;
+
+    ISwitch USBPort3S[2];
+    ISwitchVectorProperty USBPort3SP;
```
knro, line 101: Please use new style INDI properties. i.e.

INDI::PropertySwitch
INDI::PropertyText

and so forth.

knro (changes requested):

knro:
@pawel-soja Do we need to rebuild the CI image to include zeromq dependency for this to work?

cmake_modules/FindZMQ.cmake
```diff
@@ -0,0 +1,49 @@
...
+else (ZMQ_INCLUDE_DIR AND ZMQ_LIBRARIES)
+
+  find_path(ZMQ_INCLUDE_DIR
+    NAMES zmq.h
+    ${_obIncDir}
+    ${GNUWIN32_DIR}/include
+  )
+
+  find_library(ZMQ_LIBRARIES
+    NAMES libzmq3
```
knro, line 23: There is no such library. Please change this to **zmq**

AvalonService: Fixed thanks

sterne-jaeger:
Is it really a good idea to create a separate directory? Why don't simply merge it into the existing indi-avalon?

AvalonService:
We assumed this was the safest approach since the "avalon" driver is community-provided code for older hardware
while "avalonud" is managed directly by Avalon Instruments and is the official driver for our new controllers.
But obviously we are open to any suggestions...

sterne-jaeger:
> But obviously we are open to any suggestions...
As one of the developer of the StarGo driver, I would really appreciate if we use the same base directory. If there are changes necessary to merge the two of them, I'm happy to assist.

sterne-jaeger:
@knro it looks like the package libzmq3-dev is missing under Linux.

knro:
I have under 22.04, it's missing for CI images?

knro:
Will merge this now and hopefully we can fix the CI issue soon.

TallFurryMan:
I think that the github action 115 on workflow Docker in the indi core repo, triggered by my merged PR, fixed the dependency issue. We'll see the result on the next PR or commit in the indi 3rd party repo, but I'm optimistic, at least for the zmq dep.

PR #862, opened 2023-12-03
[PREVIEW] Initial focuser support added to indi_celestron_aux

ijessen:
This is functional - but testing has been limited.  Most importantly, I haven't tested with an uncalibrated focuser.

I took a slightly different approach than the focuser implementation in the core celestrongps driver - namely I adopted the interface guidance that the absolute position property value ranges from 0 to max.  I do this by applying an offset (the min value from calibration), and therefore the abs position value from the driver will be offset from the abs position displayed on the HC.  I also skipped relative positioning because I don't see a use case for it when absolute is available.

Backlash not yet implemented.

Astyle not (yet) applied.


knro:
Thanks! So what's the procedure exactly with a new uncalibrated focuser? Should there be a similar calibration workflow like in the HC driver?

ijessen:
As is there's nothing here to initiate a calibration - the interface isn't loaded unless a valid calibration is already present (you would have to use an HC or other driver to launch the calibration routine).  That said it's easy enough to add (and I see that there's an example to copy from in the dedicated celestron focuser driver in the core repo).  I'll have some time later in the week to fiddle with it.

knro:
@ijessen Did you test this? any issues?

ijessen:
> @ijessen Did you test this? any issues?

Haven't gotten around to adding in calibration support yet.  That said I've been using this in the field for a couple weeks and it's been rock solid.

knro:
Awesome, looking forward to the final driver!

knro:
@ijessen Any update? should we merge this as-is to get some community feedback? any show stoppers?

PR #864, opened 2023-12-09
No merge please - Testing Ubuntu/Fedora build.

TallFurryMan:
This MRis here  to test the build issue on Fedora and Debian. No merge please.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ INDI 3rd party drivers include all the drivers not included by default in the IN
 
 # Which drivers are in 3rd party?
 
+**This change must not be merged.**
+
 Check the [Device Driver documentation](https://www.indilib.org/devices.html) for detailed information on how to operate each device driver.
 
 All the driver documentation is community-contributed. Any updates or correction to the documentation in INDI library is greatly appreciated.


TallFurryMan:
The digest used by the Ubuntu build is:
`Digest: sha256:3ddc29f3d16d79c2a3180d421881233eb540b5248e1c9cfb78239654ea74f0d4`

The digest returned by `docker manifest inspect` for `ghcri.io/indilib/ubuntu:latest` is:
`"digest": "sha256:1edf3f189b8e223067c7daf81949063658a101ebda442b564228402c3ec1d293"`

The digest generated by the last Github Action for `ubuntu:latest` is:
`sha256:3ddc29f3d16d79c2a3180d421881233eb540b5248e1c9cfb78239654ea74f0d4`

I don't understand why the build won't pick `ubuntu:latest` as we see it from outside. I'll open a support ticket.

That said, even if the image is not the expected one, it is recent enough that builds succeed.

TallFurryMan:
So the problem lies also with `docker manifest inspect`, which does not work in all situations. The right tool for inspection is in package `docker-buildx`:

```
$ docker buildx imagetools inspect ghcr.io/indilib/ubuntu:latest
Name:      ghcr.io/indilib/ubuntu:latest
MediaType: application/vnd.docker.distribution.manifest.v2+json
Digest:    sha256:3ddc29f3d16d79c2a3180d421881233eb540b5248e1c9cfb78239654ea74f0d4
```

Here the digest is OK. That does not explain why there was a discrepancy in images at some point, as the Github Action building one image would not provide the same digest as the build executed afterwards. I suppose it is sufficient for now to say there is a non-deterministic delay between the image build and its use in MR checks.

PR #865, opened 2023-12-13
Libsvbony update to 1.12.1

jctk:
Updated SVBONY Camera SDK to Version 1.12.1

Change Log of libsvbony
 * Support SC432M camera
 * Add the SVB_BAD_PIXEL_CORRECTION_THRESHOLD parameter to set the threshold for bad pixel correction.
 * If the difference with surrounding pixels exceeds the threshold, it will be dynamically corrected.
 * It takes effect only when the SVB_BAD_PIXEL_CORRECTION_ENABLE parameter is turned on.
 * The cooling camera will no longer turn off cooling and fan when certain parameters are set.


jctk:
Hi @knro 
Could you please confirm and merge this PR?

PR #872, opened 2023-12-20
celestron-aux bugfix: track target was not updated after manual slew

### LINKED ISSUE #871: celestron-aux: track after slew bug
**Describe the bug**
When the mount is tracking a target and the user issues a move command, the tracking is paused and the mount slews as expected.  However, once the command is ended, the scope resumes tracking the *original* target.  The tracking target is not updated to reflect the mount position at the end of the move command.

There is already code in place to update the tracking target at the end of the move, but it depends on a flag `m_ManualMotionActive` being set .  This isn't happening during the move commands.

**To Reproduce**
Exact steps to reproduce the behavior.
1. Run driver
2. Track target
3. Command move away from target
4. Watch the mount slowly reacquire the original target after the move command is ended

ijessen:
fixes #871

### DIFF
--- a/indi-celestronaux/celestronaux.cpp
+++ b/indi-celestronaux/celestronaux.cpp
@@ -1036,6 +1036,7 @@ bool CelestronAUX::MoveNS(INDI_DIR_NS dir, TelescopeMotionCommand command)
     m_AxisStatus[AXIS_ALT] = (command == MOTION_START) ? SLEWING : STOPPED;
     ScopeStatus      = SLEWING_MANUAL;
     TrackState       = SCOPE_SLEWING;
+    m_ManualMotionActive |= (command == MOTION_START);
     if (command == MOTION_START)
     {
         return slewByRate(AXIS_ALT, ((m_AxisDirection[AXIS_ALT] == FORWARD) ? 1 : -1) * rate);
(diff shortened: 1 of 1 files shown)


ijessen:
> Thanks, how is the tracking performance? 

Not bad with the default PID parameters, but the polling rate absolutely needs to be higher than the default (I use 5hz).  Very sensitive to leveling and alignment.  I can take 5 second exposures pretty reliably, up to 10 seconds if the alignment is very good.

The PID is an interesting approach.  I've thought about experimenting with adding feed-forward (based on an analytical solution of the expected rates at the tracked target) but honestly it's working well enough for my purposes (short exposure EAA and planetary capture).

Only true pain point is that it's completely out to lunch before the alignment converges (at least 3 sync points), which sometimes makes alignment more difficult if tracking becomes accidentally engaged.  I haven't dug into that yet and it might be a process error on my part.

knro:
5Hz does not cause any jitter? I actually started without any PID initially, but the tracking was bad. Part of me believes that the PID loop is an overkill but I'm not sure how the hand control achieves such a good tracking performance as reported by many users. Perhaps your proposed feed-forward is a better solution? What do you mean by analytical solution exactly? 

Currently we have mount modeling which is not active by default since I found the **Nearest** algorithm to be the most reliable.

ijessen:
Honestly I haven't dug into the jitter.  Empirically my 5-10 second exposures look pretty tight, so whatever jitter is present doesn't seem significant impact.  

PID is a great fly swatter.  Maybe there are alternatives, but it almost always works to control poorly modeled systems.  

As far as feed-forward, all I was thinking was seeding the rates (i.e. PID outputs) with calculated values for the target location (i.e. derivatives of the eq->alt/az conversions) after a reset.  That would help the controller reach steady state more quickly, rather than having to come up to speed from 0.  Might also allow for a lower update rate, since convergence was one of the reasons I had to bump it up.

I haven't played around with mount modeling at all yet.  I switched to the SVD math plugin becuase - why not? - but couldn't say what difference, if any, it made.  

Any reason not to pick this PR in the meantime?  I'm absolutely going to continue to experiment with tracking (at least until I give up and get a wedge - but it's pretty good as is!), but this little PR is just addressing a single bug.

PR #875, opened 2023-12-24
update Touptek & oem lib, version: 55.24239.20231224

touptek:
Fix some model issues


knro:
THank you, can you please also update all the debian/libXXX. Just need to add an entry to the changelog file in each directory to mark a new release.

touptek:
> THank you, can you please also update all the debian/libXXX. Just need to add an entry to the changelog file in each directory to mark a new release.

ok. we will update these files from next time

PR #892, opened 2024-02-01
add missing build dependency to README

geoffrey-vl:
Encountered build issues due to the missing libboost-program-options dependency, so I figured it may help out others if this was mentioned in some place.


knro:
Wouldn't libboost-program-options-dev be sufficient?

PR #896, opened 2024-02-07
Checking if park position has reached when motors has stopped

pmneo:


indi-eqmod/eqmodbase.cpp
```diff
@@ -1065,7 +1065,22 @@ bool EQMod::ReadScopeStatus()
         {
             if (!(mount->IsRARunning()) && !(mount->IsDERunning()))
             {
-                SetParked(true);
+                currentRAEncoder = mount->GetRAEncoder();
+                currentDEEncoder = mount->GetDEEncoder();
+                parkRAEncoder    = GetAxis1Park();
+                parkDEEncoder    = GetAxis2Park();
+                if( static_cast<int32_t>(parkRAEncoder - currentRAEncoder) != 0 || static_cast<int32_t>(parkDEEncoder - currentDEEncoder) != 0 ) {
```
knro, line 1072: Use std::abs(a-b) > 0 || ...etc which is better

pmneo: Whats about 
```
if( parkRAEncoder != currentRAEncoder || parkDEEncoder != currentDEEncoder )
```
?

indi-eqmod/eqmodbase.cpp
```diff
@@ -1069,7 +1069,7 @@ bool EQMod::ReadScopeStatus()
                 currentDEEncoder = mount->GetDEEncoder();
                 parkRAEncoder    = GetAxis1Park();
                 parkDEEncoder    = GetAxis2Park();
-                if( static_cast<int32_t>(parkRAEncoder - currentRAEncoder) != 0 || static_cast<int32_t>(parkDEEncoder - currentDEEncoder) != 0 ) {
+                if( parkRAEncoder != currentRAEncoder || parkDEEncoder != currentDEEncoder ) {
```
knro, line 1072: How about we use std::abs(a - b) > threshold ? If the different is like 1 tick, should we really issue a slew command? there should be an acceptable minimal threshold. maybe 5 or 10 ticks

pmneo: Done, but why are the builds failing in a class which was not change in my branch?

pmneo:
So I can confirm, that my fix applied to night:

[2024-02-10T05:43:04.346 CET INFO ][           org.kde.kstars.indi] - EQMod Mount :  "[INFO] Motors while parking stopped, reparking mount: RA increment = 892955, DE increment = -1059060 "

CS

PR #898, opened 2024-02-18
Fix rare error Exposure failed after 3 attempts.

nouspiro:
In rare instances exposures fail. It seems like `ASIGetExpStatus` must be called in busy loop.


knro:
Thanks, let me test this locally as well.

knro:
Thank you, looks good!

PR #899, opened 2024-02-19
update  asipower properties; remove gpio

naheedsa:


knro:
Do not remove, just comment out.

PR #903, opened 2024-03-05
Add Moravian SFW driver

sWski:
Adds new driver for Standalone Filter Wheel which will be released soon. Supporting libmicam library was updated to version 0.10.0/0.9.0


knro:
Thank you, can you please attach some screenshot for the driver? link to product so we can add to documentation?

@naheedsa Please add to device documentation when ready.

sWski:
Definitely, the wheel is now anodized, my colleagues are working on the engraving, then it will be possible to take pictures. I'll let you know when the pgotos and article is ready.

PR #916, opened 2024-04-24
libqhy: fix soversion

returntoreality:
the binary file actually has the SONAME libqhyccd.so.20 and is also distributed in the official SDK as libqhyccd.so.20. This changes the SOVERSION to match that.

### DIFF
--- a/libqhy/CMakeLists.txt
+++ b/libqhy/CMakeLists.txt
@@ -3,7 +3,7 @@ project (libqhy)
 
 # QHY SDK 24.04.01
 set (LIBQHY_VERSION "24.4.1")
-set (LIBQHY_SOVERSION "24")
+set (LIBQHY_SOVERSION "20")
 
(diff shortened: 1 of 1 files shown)


knro:
That would break indi-qhy, it's always been using SO version = year in INDI at least.

returntoreality:
I created this because indi-qhy is broken because of this. During build the linker will search for libqhyccd.so, since "-lqhyccd" is given. It will find the symlink for libqhyccd.so->libqhyccd.24->libqhyccd.24.4.1 and read the file. There it will see the SONAME (can be seen with `objdump -p libqhy/libqhyccd_x86_64.bin | grep SONAME` ) and link to that SONAME. This SONAME in the master branch is `libqhyccd.so.20`. But because the libqhyccd.so is installed as libqhyccd.so.24 it is not found during runtime. If I build indi-3rdparty master I get a indi_qhy_ccd binary that was successfully linked, but then ldd shows:
```
...
libqhyccd.so.20 => not found
...
```

PR #933, opened 2024-06-16
Fix broken humidity reading after reconnect of device

peter-englmaier:
I noticed, that my cloudwatcher was showing mutliple humidity values in the parameters tab. All are zero and kstars observatory tab was not showing humidity value. This happens, when kstars disconnects and reconnects the driver. Each time the handshake method is called, adding another parameter.

This fix simply checks if the parameter already exists before adding it.

### DIFF
--- a/indi-aagcloudwatcher-ng/indi_aagcloudwatcher_ng.cpp
+++ b/indi-aagcloudwatcher-ng/indi_aagcloudwatcher_ng.cpp
@@ -71,8 +71,11 @@ bool AAGCloudWatcher::Handshake()
 
         if (m_FirmwareVersion >= 5.6)
         {
-            addParameter("WEATHER_HUMIDITY", "Relative Humidity (%)", 0, 100, 10);
-            setCriticalParameter("WEATHER_HUMIDITY");
+            // add humidity parameter, if not already present
+            if (!ParametersNP.findWidgetByName("WEATHER_HUMIDITY")) {
+                addParameter("WEATHER_HUMIDITY", "Relative Humidity (%)", 0, 100, 10);
+                setCriticalParameter("WEATHER_HUMIDITY");
+            }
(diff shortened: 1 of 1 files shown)


knro:
Thanks, it appears there is also another issue with latest firmware and the driver. Which firmware are you running? also do you have SQM?

peter-englmaier:
Yes, it has the "SQM" feature, but I have not yet tried everything. What is the other issue? 

The device was just delivered with firmware 5.89 and I have not changed it. I am having the version with RH/Pressure sensors but without the wind sensor.

knro:
is the SQM pluggable? if you unplug it and run the driver what happens?

peter-englmaier:
No, the cloudwatcher is a small white box with only two water proof connectors:

https://lunaticoastro.com/aag-cloud-watcher/
https://lunaticoastro.com/wp-content/uploads/cw-RAIN-NEW-min.png

One is for attaching the serial port, power supply, and two wires to signal the safe/unsafe switch (all with one cable). The other is for the optional wind sensor. The SQM is buildin, and I suppose it is the small round window seen at the bottom of the image. I had no time to calibrate it so far, so no idea if it works as expected.

There is another model, the "CW Hydreon version", with a different kind of rain sensor, which has 3 connectors, and the additional connector is only for that rain sensor. And, there is a "mobile" version. But I have the normal type.

My PR is for the humidity sensor, which is also buildin. It sticks out from the bottom as a black knob (maybe this is also for ambient temperature). Here you can see it:

https://lunaticoastro.com/wp-content/uploads/CloudWatcher-Relative-humidity-and-Atmospheric-pressure-sensor-v-2-02-min.jpg

If there is another bug, I can probably fix it, if I can reproduce it, but I am not aware of other problems (have the device only one week).

PR #940, opened 2024-07-09
some model bug fix & AAF & tail light on/off 

touptek:


TallFurryMan:
Failures might be that you need to provide two steps to your build system: one that builds libraries, which get installed in the system, then one that builds drivers. But you probably already know this! Thanks for your contributions!

PR #942, opened 2024-07-14
Update PlayerOneCamera SDK to v3.6.3

hiro3110i:
Quoted from SDK changelog.txt:
- Improved accuracy for short exposure time(Snap Mode)
Known Issues:On some devices(such as PI5), SDK will drop frames on Ubuntu 24.04 with linux kernel 6.7 or higher.


TallFurryMan:
Sounds good. MacOS failure is a dependency, and Fedora requires further investigation in mgen, which I'm still doing.

PR #948, opened 2024-07-22
Fix TEC temperature range initialisation

ukuehn:
Fix TEC temperature range initialisation (missing cast, missing scaling), set 0 degree as starting point instead of indicated minimum.


knro:
@touptek Please review

touptek:
@knro  Yes, this is right.

PR #965, opened 2024-08-31
fix(3rd-party): fix mgen, ahp-xc and pentax errors on fedora.

TallFurryMan:
### fix(indi-mgen): work around C++14 out-of-bounds warnings.

This is a tentative PR to fix out-of-bounds warnings emitted by C++14 when building indi-mgen.

The implementation restricts the set of function candidates the compiler selects from when using an initialiser list on IOBuffer: `IOBuffer{opcode,literal,literal...}`, and makes sure to allocate space before copying the list data.

### fix(indi-ahp-xc): silence warnings by allocating placeholders with their real size.

@iliaplatone I considered that placeholders with their real sizes instead of 1 byte everywhere wasn't that problematic. Feel free to comment.

### fix(indi-pentax): fix calloc argument order.

`calloc(sizeof,sizeof)` expects the number of items as first argument, and the size of each as second.

### DIFF
--- a/indi-ahp-xc/indi_ahp_xc.cpp
+++ b/indi-ahp-xc/indi_ahp_xc.cpp
@@ -633,43 +633,44 @@ AHP_XC::AHP_XC()
     IntegrationRequest = 0.0;
     InIntegration = false;
 
-    autocorrelationsB = static_cast<IBLOB*>(malloc(1));
-    crosscorrelationsB = static_cast<IBLOB*>(malloc(1));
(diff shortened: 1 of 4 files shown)


TallFurryMan:
Possibly the change is not minimal. Builds without warnings in `ghcr.io/indilib/fedora`, so I suppose the idea may be used elsewhere in the codebase.

TallFurryMan:
The fedora build fails because of other drivers, at least one for the same reason as this PR. I may have a look later this week-end.

PR #966, opened 2024-08-31
DRAFT: fix(libfishcamp): fix usb include location of macos.

TallFurryMan:
This PR attempts to locate `libusb.h` on MacOS, missing for libfishcamp.


knro:
Thanks! Why was the cmake module removed? there is a default one?

TallFurryMan:
Nope, there's not in the end, but I wanted to see whether the macos cmake had one. Wrong call in the end.

The location of `libusb.h` is a frequent issue, I should be able to solve this problem soon.

knro:
The issue is fixed.

PR #969, opened 2024-09-08
Fix timestamp in indi-svbony debian changelog

rkaczorek:
This fixes timestamp formatting of debian/changelog which resulted in following debian packaging error:
`dpkg-deb: error: unable to parse timestamp '': Success`


knro:
Thank you!

PR #973, opened 2024-09-10
Additional Actions

wotalota:
Extending the rolloffindi driver to be able to set additional Actions via the remote controller. The additional  actions are most likely non roof related relays.

The driver user interface always provides eight switches and labels for the switches whether or not they are used. Action on/off commands and requests status will only be sent to the controller if the controller indicates upon connection that it wants to receive them. This is to allow existing controller installs to continue working with an updated driver.

Updated Arduino examples that request the additional actions have been included.
New Arduino examples include.
- Actions used to set relays one that responds to status request and one does not.
- Open/Close roof commands that use 2 linear actuators in unison plus an action.
- Open/Close working a variable speed DC motor. 
- A standard example that is action enabled but no actions used
- The previous standard example retained to prove old controller with new driver.


knro:
Thank you! Please consider the following:
1. Use new style INDI Properties. i.e. INDI::PropertySwitch, INDI::PropertyNumber and so forth.
2. Use INDI::OutputInterface and INDI::InputInterface for any digital or analog IO sensors. You can checkout [INDI GPIO](https://github.com/indilib/indi-3rdparty/tree/master/indi-gpio) driver as an example reference.

wotalota:
Jasem,

I can look into what is involved and make the switch over to using the new properties if that is a requirement. It will take me some time to learn and retest. I think in that case it would be better to cancel the pull request until I have made the changes.

The other suggestion to use the universal roll off interface is more of a problem. The communication between the driver and the Arduino uses text strings, "open" "close". The existing Arduino implementations that are out there use code, possibly modified by the user, based on provided examples. I don't see how to keep them working if the driver side communication were to be changed.

Thank you
Tom

knro:
There shouldn't be any change on the driver side communication wise. You don't have to use INDI ROR interface at this stage, but implementing the INDI::OutputInterface and INDI::InputInterface would be highly beneficial and would even make this driver reusable by other drivers.

wotalota:
Can I just close this pull request while I working on these things?

PR #998, opened 2024-11-02
Updating the names in 2 mac libraries to fix potential linking issues

rlancaste:


knro:
Thank you! Looks like there are more MacOS errors, can you please take a look?

PR #1035, opened 2025-02-02
FIX_Reversed_Condition_Commit_fa35d75

jctk:
Correct reversed condition in function SVBONYBase::sendImage.


jctk:
Some past commits have been mixed in and will be closed to re-promote.

PR #1036, opened 2025-02-03
FIX_Reversed_Condition_in_Commit_fa35d75

jctk:
Correct reversed condition in function SVBONYBase::sendImage.


jctk:
Commit fa35d75 is the reverse of the original source code.

Original
```C
        // Send if different
        if (strcmp(bayerString, BayerT[2].text))
        {
```

Commit fa35d75
```C
        // Send if different
        if (BayerTP[2].isNameMatch(bayerString))
        {
```

PR #1037, opened 2025-02-03
For toupbase cameras build only that, which have corresponding OEM SDK library installed

vladtcvs:
* Remove `REQUIRED` field for SDK libs for toupbase cams
* Build and install only that drivers, which have SDK lib

This is useful when you build for your system and don't want to install a lot of SDK for cameras, which you don't have


knro:
Sorry for late review, but will leave the INDI Toupbase driver to be handled by @touptek

PR #1046, opened 2025-03-07
Indi ocs updates

LuckyEddie47:
Update of the indi-ocs driver.
Added OCS command protocol change on inactive weather measurements - OCS v3.08a
Fixed weather measurement polling
Fixed typo in network connection timeout value
Fixed arbitrary command debug blocks
Added command/response blocking to prevent out-of-order value updates
Refactored some if-else ladders to switch cases
Added debug option for pausing to allow remote debugger to attach


LuckyEddie47:
See Issue #1047

PR #1061, opened 2025-04-02
update Touptek & oem lib, version: 57.28165.20250401

touptek:


knro:
@touptek Please rebase first then submit again.

touptek:
> @touptek Please rebase first then submit again.

ok, we will submit a new & clean version 3 or 5 days later

PR #1062, opened 2025-04-02
Bresser Exos 2 - Support for x86/x64 architectures and Sync function from planetarium appplications

damianofasoli:
Hello 

I have already committed these updates

into my repository  
https://github.com/damianofasoli/indi-bresserexos2

and @kneo  repository
https://github.com/kneo/indi-bresserexos2

is aligned

There is a fix for x86/x64 architectures.
Without it the driver works with ARM architectures but with x86/x64 architectures RA and DEC coordinates are not sent to the driver at all

There is also a different Sync function which does not talk at all with the HandBox, but keeps coordinate variations  (when the mount is moving) inside  the driver, making possible to Sync from planetarium applications (Ekos/Kstars for example) and to do Plate Solving 
Coordinates variations produced in this way are not sent to the HandBox. Often the two systems of coordinates are already different, even without this update

Unfortunately, I do not have more an Exos 2 mount to develop and test further updates, but these updates was developed and tested when I owned the mount


knro:
Thank you. It's not clear why this change would only affect a specific architecture?

damianofasoli:
Ok, these are two different issues. For x86/x64 there was a bug due to
signed/unsigned char causing the mount not working at all. The other issue
Is a general improvement for all architecturee

PR #1069, opened 2025-04-11
feat(atik): add gain/offset control outside of CCD_CONTROLS.

TallFurryMan:
This MR adds two properties to the Atik driver, `CCD_GAIN` and `CCD_OFFSET`, allowing adjustment of gain and offset in the same way as `CCD_CONTROLS[0]` and `CCD_CONTROLS[1]`. These properties should behave like those of the QHY driver.

I don't have a APX at hand, I expect the properties to work identically on Horizon, Horizon2 and APX, but I don't iknow if they will be setup properly for APX.

### DIFF
--- a/indi-atik/atik_ccd.cpp
+++ b/indi-atik/atik_ccd.cpp
@@ -138,6 +138,15 @@ bool ATIKCCD::initProperties()
     IUFillText(&VersionInfoS[VERSION_FIRMWARE], "VERSION_FIRMWARE", "Firmware", "Unknown");
     IUFillTextVector(&VersionInfoSP, VersionInfoS, 2, getDeviceName(), "VERSION", "Version", INFO_TAB, IP_RO, 60, IPS_IDLE);
 
+    // Direct Gain Control
+    IUFillNumber(&GainN[0], "GAIN", "Gain", "%.f", 0, 60, 5, 30);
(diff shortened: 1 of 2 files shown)


TallFurryMan:
Please wait a little, I will test this on my Horizon2.

knro:
Why do you need to add these properties?

TallFurryMan:
> Why do you need to add these properties?

In KStars, Ekos does not write the gain and offset properties the way they are written in `ekos/capture/captureprocess.cpp`. Ekos writes `CCD_GAIN` and `CCD_OFFSET`, or falls back to `CCD_CONTROLS["Gain"]` and `CCD_CONTROLS["Offset"]` if the former cannot be found.

I parsed a few other drivers to look at how gain and offset were managed, and I took example on the QHY driver, which provides `CCD_GAIN` and `CCD_OFFSET` directly. I don't want to break the existing implementation of `CCD_CONTROLS["CCD_GAIN"]` and `CCD_CONTROLS["CCD_OFFSET"]`, so my proposal attempts to handle both. Of course, having two effectively identical properties with different names and locations in the property tree isn't that ideal.

TallFurryMan:
FYI, the PR works as intended. Ekos uses CCD_GAIN and CCD_OFFSET as first source of gain/offset for Atik Horizon cameras.

TallFurryMan:
However there is a glitch with CCD_CONTROLS. `CCD_CONTROLS[Gain]` and `CCD_CONTROLS[Offset]` are not forwarded to the client when CCD_GAIN and CCD_OFFSET change, and there is an unexpected `CCD_CONTROLS` property with value `GO Controls` present. I'll investigate.

PR #1071, opened 2025-04-15
Attempted solution to compilation issue from uint8_t removal from std…

rlancaste:
…::char_traits in LLVM 19

the template for char_traits<unsigned char> was removed from the standard.   This made the macOS build fail once Xcode was updated to 16.3.

Solution was recommended here: https://gitlab.kitware.com/vtk/vtk/-/issues/19586


knro:
@rlancaste Do you know why it keeps failing on MacOS still?

PR #1072, opened 2025-04-15
Attempted solution to compilation issue from uint8_t removal from std…

rlancaste:
…::char_traits in LLVM 19

the template for char_traits was removed from the standard. This made the macOS build fail once Xcode was updated to 16.3.

Solution was recommended here: https://gitlab.kitware.com/vtk/vtk/-/issues/19586

### DIFF
--- a/indi-qhy/qhy_focuser.h
+++ b/indi-qhy/qhy_focuser.h
@@ -70,7 +70,7 @@ class QFocuser : public INDI::Focuser
         double targetPos{ 0 };
         bool isReboot = false;
 
-        uint8_t buff[USB_CDC_RX_LEN];
+        char buff[USB_CDC_RX_LEN];
 
         int32_t cmd_version;
         int32_t cmd_version_board;


rlancaste:
No, I didn’t actually change them, those libraries install names were changed by my macro in the repo that automatically scans libraries for issues with the install name of the libraries and updates them since when you build on macOS with bad install names it causes serious and unexpected problems at the linking step.  We have many libraries in the 3rd party repo and they are always getting updated by various people with different system configurations.  The macro fixes the problem.  I added it several years ago.  Without this script, the build broke for macOS on a very regular basis and we had to go in and manually change the install names after every update people made.

TallFurryMan:
Umm, I'm very sorry about this, I wasn't aware. I'll fix my import script, thanks!

rlancaste:
Hi Eric,  It's no problem, many of the libraries have incorrect paths for their "install names"  when we get them.  Usually it is because people build on their computer where their libraries could be in homebrew, macports, the same folder as the build files, or somewhere else.  It does work on their computers, but then when they send that binary library to the repo and someone tries to build the program that relies on that library, it finds the library and then doesn't link properly, Sometimes it looks like it built perfectly, but then when they go to run it, it fails to load the library.  This is particularly important for KStars and similar programs that bundle their libraries in the App bundle for distribution.  I usually find that an install name based on rpaths works  best, since then, the executable can specify the rpath based on how it is "installed" and it will find the library and it will still load.  Just take a look at my macro at the top of the main cmakelists in 3rdparty and you will see how it works.

PR #1078, opened 2025-04-25
spec_213: update spec files for version 2.1.3

xsnrg:


knro:
@xsnrg It's failing due to gtest not compatible with C++17?

xsnrg:
The build in github was probably failing before this PR too then. I only changed version numbers. ahp-gt was moved into the eqmod package so it needed to be removed from the dependent packages at build time. This was causing the build to fail for eqmod. I checked the change in a branch build in copr: https://copr.fedorainfracloud.org/coprs/xsnrg/indi-3rdparty-bleeding/build/8965588/

I may be able to look at the github build later.

PR #1088, opened 2025-05-12
Gain fix and update MI library

sWski:
- fix small bug where some cameras did not correctly update gain
- new library with support for C1+/C2-46000 cameras


knro:
Thank you!

PR #1089, opened 2025-05-15
Add missing cstdint inlcude for uint32_t

darekp:
The camera_status.hpp file uses uint32_t, which is defined in the <cstdint> header.

Including this header resolves a compilation error.


knro:
Thank you!

PR #1092, opened 2025-05-19
env for disabling udev for flatpak

s-pouya-sh:


knro:
```
option(INDI_INSTALL_UDEV_RULES "Install UDEV rules" On)
```
Should be added for each individual project separately because each project can be built on its own.

PR #1122, opened 2025-07-18
adding driver declaration

djibb:
CCDCIEL needs driver declaration in order to list SPOX module


indi-shelyak/indi_shelyakspox_spectrograph.cpp
```diff
@@ -410,4 +410,12 @@ bool ShelyakSpox::calibrationUnitCommand(char command, char parameter)
...
 
-//other lamps
+bool ShelyakSpox::saveConfigItems(FILE *fp)
+{
+    // Call the parent's saveConfigItems
+    INDI::DefaultDevice::saveConfigItems(fp);
+
+
+    return true;
+}
```
knro, line 420: Why do you need to explicitly called here? It should be managed already by INDI::DefaultDevice

djibb: Done.

PR #1137, opened 2025-08-10
Support for SHT30 sensor added to weatherradio + driver bugfix

sterne-jaeger:
Support added for the SHT30 temperature and humidity sensor to weatherradio.


knro:
@sterne-jaeger is this still in Draft?

PR #1146, opened 2025-08-21
Migrate driver to new INDI properties

knro:
Migrate driver to new properties and save all possible configs in the configuration file.


knro:
@gy-lehel  Can you please check this PR and see if it works well for your setup?

gy-lehel:
Tail light setting for the 585 has disappeared. CG setting is saved, and restored. Camera is dysfunctional though, can not even set the cooler on/off with the current touptek firmware.

gy-lehel:
CG setting is restored on control panel, but I am not able to validate the status of the camera, as with the firmware+driver combo in 2.1.5 the 585 family of touptek cameras is completely disfunctional, it is not possible to take any exposures.

knro:
I can't reproduce the taillight issue. Ok will merge this for now and let's see other users feedback

PR #1154, opened 2025-08-26
Update libASICamera2 to version 1.39 from ASI SDK

geeksville:
Hi ya'll,

Not sure if this is useful for you but while debugging another problem with a new ASI camera (see below) I tried updating the ASI Camera SDK to version 1.39 (from https://www.zwoastro.com/software/).  It was released on 2025-08-21.  I'm new to INDI so someone who knows how you like to pull in SDK lib blobs should probably review this.

btw: The 'real' problem I set out to solve was: Figure out why my ASI MC2600 Duo camera fails when trying to open the primary camera (only the internal guide camera succeeds).  I think I'm close to figuring out the root cause and a fix.  If I make progress I'll send in a PR for that.


knro:
Thank you. You need to run dos2unix on the header file. The debian/indi-asi/changelog needs to be also updated. So with this SDK, your issue is resolved? I think many are using MC2600 Duo with the existing driver/SDK without any issues.

geeksville:
And I didn't test on mac, I kinda just assumed the libASICamera2.dylib from ASI should be slammed in as libASICamera2.bin for that arch.

PR #1166, opened 2025-09-05
Bugfix workerExposure function

fgranjon42:
Fix workerEpxosure function , rewrite loop

When exposureTimeLeft > 0.6 
Refresh indi client and call  POAImageReady all 0.5 seconds 
When exposureTime < 0.6 
Only call POAImageReady all 1000 microseconds


knro:
Thank you @fgranjon42 !

@hiro3110i Can you please check?

PR #1167, opened 2025-09-05
Draft: Toupcam, restore taillight setting and resolution on startup

gy-lehel:
Add code to restore tail light setting on driver start-up. Forcibly cycle resolution setting to work around touptek SDK bug blocking image download.

### DIFF
--- a/indi-toupbase/indi_toupbase.cpp
+++ b/indi-toupbase/indi_toupbase.cpp
@@ -505,6 +505,20 @@ bool ToupBase::Connect()
         int taillight = 0;
         HRESULT rc = FP(get_Option(m_Handle, CP(OPTION_TAILLIGHT), &taillight));
         m_SupportTailLight = SUCCEEDED(rc) ? true : false;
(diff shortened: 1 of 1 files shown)


gy-lehel:
Tested on aarch64 and x86_64. Restores taillight setting in-camera, resolution cycling solves image download issues, conversion gain setting is also restored in-camera.

knro:
@HLex79 Can you please test this if you can?

gy-lehel:
Do you have any experience with compiling, and installing software on linux from source code?

knro:
I applied the patch on @HLex79 machine, but the same issue persisted. We had to switch the resolution to get it working. The put_eSize apparently didn't resolve the issue.

gy-lehel:
BTW yesterday evening my 585 failed big time with this patch, and manual setting, too. changed filters, and had to refocus, and when started looping in order to set it with a bathinov mask, it failed after 3-5 images.

PR #1173, opened 2025-09-14
Split up toubase into individual manufacturer libraries.

gy-lehel:
Currently building any of the touptek rebrands requires all the different SDK-s and drivers to be built. In this PR I am trying to split up that folder into a per-manufacturer based smaller ones.


gy-lehel:
Then I'll try a different approach, but it will take some time to implement it.

gy-lehel:
Please have a look, I split the building on cmake level, creating separate build targets for the different rebrands, while maintaining the single source directory. This allows to build and install only the sdk and driver only for the selected brand, instead of all of them.

PR #1191, opened 2025-10-05
Restore touptek tail light and conversion gain setting on driver start-up

gy-lehel:
Current implementation restores values only in the GUI control elements.

If conversion gain setting retrieved from the camera
differs from the one configured by the user, set the configured value
in the camera on driver start-up.
If tail light setting retrieved from the camera
differs from the one configured by the user, set the configured value
in the camera on driver start-up.

Add additional error handlers to signal the user if restoring/changing of these values failed.

### DIFF
--- a/indi-toupbase/indi_toupbase.cpp
+++ b/indi-toupbase/indi_toupbase.cpp
@@ -21,6 +21,7 @@
 
 #include "indi_toupbase.h"
 #include "config.h"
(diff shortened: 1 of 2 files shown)


gy-lehel:
Then we should also fix the differences between the astyle, and clang-tidy settings, as the file is formatted with the project-level clang-tidy settings found in the repo root.

PR #1224, opened 2025-12-15
Adapt driver to expect Now from base class and convert to J2000 for telescope

jrrk2:
Fix coordinate system mismatch causing 18.9 arcmin pointing error

The Origin telescope uses J2000 coordinates internally, but INDI's
Telescope base class only exposes EQUATORIAL_EOD_COORD (JNow/EOD).

This creates an absurd conversion chain:
1. Client converts J2000 → JNow (for INDI)
2. Driver converts JNow → J2000 (for mount)
3. Driver converts J2000 → JNow (reporting to INDI)
4. Client converts JNow → J2000 (for display)

This is a fundamental design flaw in INDI that forces all modern
mounts (which use J2000) to perform unnecessary conversions. The
proper solution would be for INDI to support J2000 coordinates
natively, but for backward compatibility we implement the conversion
in the driver.

The libnova library handles the precession calculations correctly.

1. Star Chart (Stellarium/etc)
   ↓
   M31 J2000: 00h 42m 44s, +41° 16' 09"
   ↓
2. CCDciel
   ↓ [CONVERTS J2000 → JNow]
   ↓
   M31 JNow: 00h 44m 10s, +41° 24' 56"
   ↓
3. INDI Protocol (EQUATORIAL_EOD_COORD)
   ↓
   Sends JNow: 00h 44m 10s, +41° 24' 56"
   ↓
4. Your Origin Driver
   ↓ [CONVERTS JNow → J2000]
   ↓
   Back to J2000: 00h 42m 44s, +41° 16' 09"
   ↓
5. Origin Mount (which uses J2000)
   ↓ [CONVERTS J2000 → Alt/Az for motors]
   ↓
   Calculates motor positions based on J2000
   ↓
6. Mount reports position
   ↓
   Reports J2000: 00h 42m 44s, +41° 16' 09"
   ↓
7. Your Origin Driver  
   ↓ [CONVERTS J2000 → JNow]
   ↓
   JNow: 00h 44m 10s, +41° 24' 56"
   ↓
8. INDI Protocol
   ↓
9. CCDciel
   ↓ [CONVERTS JNow → J2000 for display]
   ↓
   Shows: 00h 42m 44s, +41° 16' 09"

In addition the Gain/ISO setting syntax was changed to make it functional in kstars.

### DIFF
--- a/indi-celestron-origin/indi_origin.cpp
+++ b/indi-celestron-origin/indi_origin.cpp
@@ -640,10 +640,10 @@ bool OriginCamera::initProperties()
     PrimaryCCD.setMinMaxStep("CCD_EXPOSURE", "CCD_EXPOSURE_VALUE", 0.000001, 3600, 0.000001, false);
     
     // Gain/ISO property - Origin supports 0-1600
-    // Match the simulator pattern exactly
-    GainNP[GAIN].fill("GAIN", "value", "%.f", 0, 1600, 1, 100);
+    GainNP[GAIN].fill("GAIN", "Gain", "%.0f", 100, 1600, 100, 200);
     //                                          ^  ^     ^  ^
     //                                        min max  step default
+    GainNP.fill(getDeviceName(), "CCD_GAIN", "Gain", MAIN_CONTROL_TAB, IP_RW, 60, IPS_IDLE);
     
     // Preview/Full mode property
     StreamSP[STREAM_PREVIEW].fill("PREVIEW", "Preview (fast)", ISS_OFF);


knro:
You can define ```EQUATORIAL_COORD``` in your driver which is J2000 based coords. You would still need to convert J2000 to EOD when calling ```NewRaDec(double ra, double dec)``` since that function expect JNow coordinates. However, you will receive J2000 requests and can pass them directly to the driver.

jrrk2:
I understand that I can expose whatever properties I choose. However I do not find a frequent pattern of J2000 usage in the base class or other drivers so have no reason to believe that this path would be well tested in kstars or CCDciel. So I opted for this method which would appear to be compatible with the greater body of code, and as you say I need to do the conversion anyway. I tested the fix on my telescope and both of these programs last night and it is a good improvement in slewing accuracy. I am reluctant to start removing _EOD_ properties from the Base Class at runtime and providing both options would provide code paths which are rarely used and would reduce code coverage and maintainability.

PR #1225, opened 2025-12-19
ASI EAF: add serial-number–based naming and `ZWONicknames.xml` support

tbowers7:
### Description

This PR adds persistent nickname support for ZWO EAF focusers, allowing multiple identical devices to be uniquely identified using their hardware serial numbers.

Nicknames are stored in `~/.indi/ZWONicknames.xml` and are applied automatically at driver startup when a matching serial number is found.

This brings the ASI EAF focuser driver in line with the ASI CCD driver by adding support for SDK-provided serial numbers and the existing `ZWONicknames.xml` nickname mechanism.

### Motivation

When multiple ZWO EAF focusers are connected, they currently appear with identical or ambiguously numbered device names, making it difficult to reliably associate a focuser with a specific telescope or optical train.

ASI cameras already support serial-number-based nicknames, but EAFs lacked equivalent functionality. This PR brings feature parity and improves multi-device usability.

### What this PR does

* Queries the ZWO EAF hardware serial number during device discovery.
* Exposes the serial number as a read-only INFO property.
* Adds a writable Nickname property for configuring persistent device names.
* Stores nicknames in `~/.indi/ZWONicknames.xml`, using the same XML format and parsing logic as `asi_ccd.cpp`.
* Applies nicknames at driver startup when a matching serial number is present.
* Allows nickname changes at runtime, with changes taking effect after driver restart.

### Compatibility

* No behavior changes unless a nickname is explicitly defined.
* No new external dependencies are introduced.
* Uses the existing LilXML utilities already used by ASI CCD drivers.
* Compatible with older INDI SDK versions.
* Existing configurations and setups remain unaffected.

### Testing

* Tested with a single ZWO EAF: default naming unchanged when no nickname is defined.
* Tested with multiple identical EAFs: serial numbers correctly detected and nicknames applied
* Verified nickname persistence across driver restarts.

This change mirrors the established behavior of the ASI CCD driver and improves usability for multi-focuser configurations.

### DIFF
--- a/indi-asi/README.md
+++ b/indi-asi/README.md
@@ -50,6 +50,31 @@ changed, a flushing mechanism is employed. It looks OK, but if you
 find any problem with parameters changes not being immediately applied
 please report.
 
+NICKNAMES
+
+The ASI SDK exposes device serial numbers for at least CCDs and EAFs.
+You may associate nicknames with specific serial numbers to make, *e.g.*,
+setups with multiple identical devices reliably associate the same name
+with the same device across restarts.
+
+Nicknames are stored in an xml-format file in a format like the below.
+The hard-wired location for this file is ``~/.indi/ZWONicknames.xml``.
+Nicknames are assoicated with the serial number of the camera, and are
+entered/changed with the NicknameTP text property. Since the device-name
+can't be changed once the driver is running, changes to nicknames can
+only take effect at the next INDI startup.
+
+You may mix CCD and EAF nicknames in the same file, using the format:
+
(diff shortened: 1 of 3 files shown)


tbowers7:
> @tbowers7 perhaps a more generic framework that can be used for all drivers can be devised?

@knro , yes, that is the intention of issue #1226 , with the present PR being a stop-gap for my current use case.  My hope was that someone with more familiarity with C++ would be able to devise that more generic framework.

PR #1237, opened 2026-01-02
Align with upstream libraries

iliaplatone:


knro:
Thank you @iliaplatone . Please check the build failures.

iliaplatone:
> @iliaplatone Any update?

Yes, it seems that debian main has the latest libraries, Ubuntu not instead, but there is a ppa with them ppa:iliaplatone/ppa

PR #1244, opened 2026-01-11
Add a new driver: indi-weather-mqtt

rkaczorek:
This commit adds a new INDI driver for weather monitoring from an arbitrary MQTT source.
This driver is superior to any existing weather driver as it can be fed with data from virtually any source through local or public MQTT broker. If your weather station can publish to MQTT topics, you can use it right away by subscribing to the weather topics. Otherwise you can parse ANY weather data source with a middleware (eg. node-red) and publish it to a MQTT broker. Then you subscribe to the weather topics and enjoy!


knro:
@rkaczorek Any update on the documentation? it's required now for any submitted driver.

rkaczorek:
I will look into it this week.

rkaczorek:
Ready indilib/drivers-docs/pull/15

PR #1255, opened 2026-01-27
Tweak GAIN_STEPS calculation in celestronaux.h

jochym:
Updated GAIN_STEPS to a new calculation based on geometric analysis and testing.

Rate based on geometric analysis and testing against SkySafari 7, simulator and real mount:
// Logical unit for guiding commands is 1/1024 arcsec/sec.
// Steps per arcsecond = 16777216 / (360 * 3600) = 16777216 / 1296000
// Scaling Factor (Units -> Steps/sec) = 1024 * (1296000 / 16777216)
// Factor = (1024 * 1296000) / 16777216  = 1327104000 / 16777216
 // Simplified Rational Factor = 10125 / 128
 // Steps/sec = Value * (10125 / 128) = Value * 79.1015625

The previous approximate measured value (80) - led to a small drift of ~10'/h
The correction makes this go away. 

This is tested agains SS7, unfortunately the wheather prohibits teting ATM - thus WIP designator.

I will report as soon as I get the data.

### DIFF
--- a/indi-celestronaux/celestronaux.h
+++ b/indi-celestronaux/celestronaux.h
@@ -482,8 +482,17 @@ class CelestronAUX :
         static constexpr double STEPS_PER_HOUR {STEPS_PER_REVOLUTION / 24.0};
         static constexpr double HOURS_PER_STEP {24.0 / STEPS_PER_REVOLUTION};
 
-        // Measured rate that would result in 1 step/sec
-        static constexpr uint32_t GAIN_STEPS {80};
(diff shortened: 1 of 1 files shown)


jochym:
@knro Last night I have an opportunity to test it and the real hardware confirms much smaller drifts with the changed factor. Essentially it is ~8'/h without and <1'/h with the correction. It was a very windy night, so I could not really test the final accuracy but the improvement is obvious. Please merge this PR.

PR #1289, opened 2026-03-20
fix(touptek): flush on timeout, fix bit depth and initial format, improve timeout management.

TallFurryMan:
This MR fixes a frame corruption issue when starting the Touptek driver.

An incorrect initial frame format requires the end-user to switch between RGB and RAW16 to restore functionality.

An incorrect bit depth mapping causes frame corruption until the firmware fixes the parameter by itself.

These fixes resolved all the timeout issues I had with my G3M662, even for frames under one second.


knro:
@TallFurryMan Any updates?

TallFurryMan:
Nope, couldn't continue on the streaming problem, which takes root deeper in the driver. However, for single frame capture, this PR works perfectly. No issues whatsoever with the automation of the G3M662 in the obs since deployment.

Issue #1239, opened 2026-01-05
Celestron_aux with Advanced GT WiFi is not moving on pulse guiding commands


svalat:
**Describe the bug**
Problem of using guiding with the driver `celestron_aux` on a `Celestron Advanced GT` via the `WiFi celestron dongle`.

Problem seen on the Raspberry PI 5 controling the setup or directly controling the setup from my PC.

**To Reproduce**
Exact steps to reproduce the behavior.
1. Start indiserver with the following config
2. Connecting KStars / Ekos to the server
3. Making manual Focus
4. Making Polar alignment with Ekos assistance.
5. Trying to enable Guiding via `ALTAIR GPCAM3 462M2` configured as `OAG` on top of the `QSI 683` behind a `Celestron C8` + `Reducer 0.63x`.
6. See the stars in ALTAIR camera.
7. Start guiding
8. Calibration sometime fail (not moving enougth the star), sometime succeed due to noise with wrong pattern in `calibration view` with not a cross as expected but a single line following the noise so both axis AD & DEC are mapped on the ~ same direction.
9. Try to manually send `guiding` pulse via the `manual pulse window` => not moving the star while refreshing the camera.

```sh
start indi_qsi_ccd -n "QSI 683-wsg8"
start indi_celestron_aux -n "Celestron Advanced GT WiFi"
start indi_joystick
start indi_altaircam_ccd -n "ALTAIR GPCAM3 462M2"
```

**Expected behavior**
The mount should move when sending pulse either via : 

- automatically when guiding calibration is ongoing
- via pulse windows in guiding tab of Ekos
- via `guiding` tab in Configuration tab of the mount.

Not working in the last 3 configurations.

Ekos report in the view that is sends the event, but no action is performed by the mount (or not visible on camera).

**Screenshot**

<img width="953" height="443" alt="Image" src="https://github.com/user-attachments/assets/02fbd822-97d5-4188-a5eb-bcc96b7eb07e" />

<img width="805" height="291" alt="Image" src="https://github.com/user-attachments/assets/1733da1f-0aa7-47b8-a626-bebab932b5c4" />

**Also tried**

Using PHD2 as guider via INDI to connect the camera & the mount => same issue, calibration fail due to no move over west/east/north/south during calibration to obtain the cross to get the angle of the camera.

**Working situation so hardware is OK**

The setup is working if I'm using:
1. Guiding camera through EKOS under the Raspberry PI 5
[...]

PR #1297, opened 2026-04-07
celestronaux: guide rate drift autoguiding option to support CG5 + celestron dongle wifi

svalat:
To the indi team : Feel free to edit and comment the patch, this is a proposal.

The problem
-------------------

As discussed in #1239 I had some issues with the celetronaux driver with my **Celestron CG5** + **Celestron SkyPortal Dongle WiFi** to enable autoguiding.

The **celestronaux** driver currently send the `PULSE` command to the mount which are not taken in account by my setup.

At the opposite, running with **Celestron CPWI** + **PH2** shows that the `PULSE` command are replaced by commands updating the `guide rate` (`MC_SET_POS_GUIDERATE` and `MC_SET_NEG_GUIDERATE`) by shifting it about a percentage of the sideral speed (80% by default).

All the analysis of the problem can be fount in #1239 by tracking the packets send to the mount by **CPWI**.

By the way, the **EQMod** driver also implement this approach.

This patch
---------------

I implemented the solution in the CelestronAux driver as an optional feature which needs to be enabled in the contrôle panel of the mount named **PULSE_MODE** and which offer two values :

-`PUSLE` : the default, corresponding the current implementation, still valid and in place.
-`GUIDE RATE` to be enabled on demand to enable the **guide rate drife** approach of **EQMod** and **CPWI**.

impact to the user
---------------------------

No impact by default, if the mount do not react to `PULSE` mode the user has to enable the `GUIDE RATE` mode.

The selected mode is saved as other values to enabled again on restart.

The implementation
------------------------------

I added a few states to the driver to track the status of the guiding and to restore (`restoreTrackingRateMode()`) it to its normal value when the pulse has been done.

The delay is tracked by two new timers, one for each axis. Note that if a pulse ingoing and a new one arrive, I extend the duration of the pulse.

I added a check about the hemisphere (` if (isNorthHemisphere() == false)`), to be checked. At this location I'm not sure of the sign of the existing line : 

```cpp
//for east
guidePulse(AXIS_RA, ms, -rate);
//for west
guidePulse(AXIS_RA, ms, rate);
```

But to reproduce **CPWI** behavior I need to invert this sign (I currently let in this patch the old way).
In practice this is not a big issue because PHD2 auto adap to the moves when it trains itself.

To look at
--------------

I arbitary limit the pulse time to 10 seconds (10000 ms) as there was already a limit on PULSE probably due to hardware limitations. In practice this limit is not required but avoids possible issues if for some reason the user set manually an hour which might be not desired because we cannot stop it once it is triggered.

Validation
--------------

I passed 3 full nights with a draft of this patch + 1 full night with this cleaned one.

The validation setup is : 

- Telescope : Celestron C8
- Mount : Celestron CG5
- Contrôle : Via SkyPortal Celestron WiFi dongle
- Tracking cam : Altaïr GPCAM 462M2 with OAG
- Tracking: PHD2 (version 2.6.14)
- OS: Gentoo
- Version: latest (on ARM on Raspberry-PI).
- Indlib : 2.1.7
- Kstars : 3.8.0


knro:
@svalat Any chance to test this against latest INDI v2.2.0 just to rule out any issues?

svalat:
Hi, Yes I will upgrade and make the test next clear night. I will not be available next week so it will be after. I will report as soon as possible.

knro:
@svalat I'll merge this for now and ask for users feedback. It would be great if you can let me know how your testing went. But more feedback from users would be great as well.

PR #1305, opened 2026-04-17
Use Nicknames for ASI & QHY CCDs

tbowers7:
Use the new generalized nickname scheme in INDI core (https://github.com/indilib/indi/pull/2343) for ZWO/ASI and QHY CCD cameras.  The this basically involves calling the `DefaultDevice` function `setDeviceNicknameFromId()`, and the functionality in the core library takes it from there.

#### Example

Here is an example `Nicknames.xml` file from a setup that uses four different cameras and three focusers:

```
<?xml version="1.0"?>
<INDINicknames>
  <device name="ZWO EAF">
    <nickname identifier="0F23C2003020F901">SM</nickname>
    <nickname identifier="0F23218028DBC1A1">WR</nickname>
    <nickname identifier="6000206EF19EE420">WB</nickname>
  </device>
  <device name="ZWO CCD">
    <nickname identifier="081a0f011a020900">ASI6200MM-Pro WB</nickname>
    <nickname identifier="212b810522020900">ASI6200MM-Pro WR</nickname>
    <nickname identifier="091c920d2d010900">ASI183MM-Pro SM</nickname>
  </device>
  <device name="QHY CCD">
    <nickname identifier="QHY294PROC-4ce22a4b811ae7175">QHY294C UW</nickname>
  </device>
</INDINicknames>
```

When these are queried using the command `indi_getprop | awk -F'.' '{print $1}' | sort -u1`, the resulting list is:
```
QHY CCD QHY294C UW
ZWO CCD ASI183MM-Pro SM
ZWO CCD ASI6200MM-Pro WB
ZWO CCD ASI6200MM-Pro WR
ZWO EAF SM
ZWO EAF WB
ZWO EAF WR
```

### DIFF
--- a/indi-asi/README.md
+++ b/indi-asi/README.md
@@ -1,8 +1,12 @@
+ZWO Optics ASI CCD Driver
+=========================
+
 This is the INDI driver for the ZWO Optics ASI cameras. It was tested
 with the ASI120MC and the ASI120MM but, hopefully, should work with
 other cameras from ZWO Optical too.
 
 COMPILING
+---------
 
 Go to the directory where  you unpacked indi_asi sources and do:
 
@@ -16,12 +20,14 @@ make
 should build the indi_asicam executable.
 
 RUNNING
+-------
 
 The Driver can run multiple devices if required, to run the driver:
 
 `indiserver -v indi_asi_ccd`
 
 AVAILABLE CONTROLS
+------------------
 
 You can set the exposure time and the gain (in the range between 1 and
 100) You can also change the USB Bandwidth control. It defaults to -1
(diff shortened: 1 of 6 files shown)


knro:
Thanks! Looks good to me. @perfinion Any comments?

tbowers7:
One thought/question for @knro before approval/merge: Should we extend the nickname support to other QHY or ASI devices?  I don't have any such to test on, but the addition of nickname support should be straightforward.

perfinion:
Awesome, thanks for the cleanup! It looks good, my only question is about handling the migration. Long term we definitely want it to look like this, but should we have at least one release where it reads the old ZWONicknames and then uses that to set the new generic one?
I did not handle migration for the EAF because it turned out there was a bug so no one would have had the nicknames, but this camera one has been working so maybe we want to migate things nicer?

knro:
Not sure how many people were using the old scheme, but migration would be OK, so we can simply ask them to recreate again.

PR #1314, opened 2026-04-30
indi-toupbase: fix streaming with binning

cfuture81:
Fix frame size mismatch when streaming with binning > 1x1 on ToupTek cameras.

**Problem**: `allocateFrameBuffer()` always used full sensor resolution for buffer 
size and `Streamer->setSize()`, ignoring the current binning. This caused streaming 
to deliver wrong-sized frames when binning was active. Additionally, binning changes 
from other modules (e.g. capture at 2x2) persisted into streaming even when the 
client requested 1x1.

**Fix**:
- `allocateFrameBuffer()` now divides by BinX/BinY when calculating dimensions
- `StartStreaming()` re-applies current binning to hardware before entering video mode

**Tested with**: ToupTek GPM174M at 1x1 and 2x2 binning in streaming mode.

### DIFF
--- a/indi-toupbase/indi_toupbase.cpp
+++ b/indi-toupbase/indi_toupbase.cpp
@@ -832,19 +832,24 @@ void ToupBase::setupParams()
 
 void ToupBase::allocateFrameBuffer()
 {
+    uint32_t binX = PrimaryCCD.getBinX();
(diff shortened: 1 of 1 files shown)


TallFurryMan:
I think you nailed the issue I was left with in 99e26e11, thanks! 👍

PR #1320, opened 2026-05-11
alignment: migrate to TransformCelestialToTelescopeJD / TransformTelescopeToCelestialJD

ckemper67:
## Summary

- Migrate all deprecated `TransformCelestialToTelescope` / `TransformTelescopeToCelestial` call sites in `indi-celestronaux` and `indi-eqmod` to the JD-explicit variants introduced in indilib#2388
- In `celestronaux.cpp`, hoist `ln_get_julian_from_sys()` at the top of each affected scope so multiple call sites in `Goto()`, `mountToSkyCoords()`, and the TimerHit tracking block share a single clock read per tick
- In `eqmodbase.cpp`, reuse the existing `juliandate` variable already in scope for `TransformTelescopeToCelestialJD`; call `ln_get_julian_from_sys()` inline for `TransformCelestialToTelescopeJD`
- Fix `test_eqmod.cpp`: `eqmod.horizon` was changed from a raw pointer to a `unique_ptr`, requiring `.get()` when assigning to a raw pointer local

## Test plan

- [ ] `indi_celestron_aux` builds without deprecation warnings
- [ ] `indi_eqmod_telescope` builds without deprecation warnings
- [ ] `test_eqmod` encoder/hemisphere tests pass (5/8; 3 scope-limits tests require installed XML data files and were already failing before this change)

### DIFF
--- a/indi-celestronaux/celestronaux.cpp
+++ b/indi-celestronaux/celestronaux.cpp
@@ -1269,7 +1269,7 @@ INDI::IHorizontalCoordinates CelestronAUX::AltAzFromRaDec(double ra, double dec,
     TelescopeDirectionVector TDV;
     INDI::IHorizontalCoordinates AltAz;
 
-    if (TransformCelestialToTelescope(ra, dec, ts, TDV))
+    if (TransformCelestialToTelescopeJD(ra, dec, ln_get_julian_from_sys() + ts, TDV))
         // The alignment subsystem has successfully transformed my coordinate
         AltitudeAzimuthFromTelescopeDirectionVector(TDV, AltAz);
     else
@@ -1745,6 +1745,7 @@ bool CelestronAUX::Goto(double ra, double dec)
     }
 
     uint32_t axis1Steps {0}, axis2Steps {0};
+    double JDnow {ln_get_julian_from_sys()};
 
     TelescopeDirectionVector TDV;
     INDI::IEquatorialCoordinates MountRADE { ra, dec };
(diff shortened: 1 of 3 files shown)


jochym:
Unfortunately recent changes made the celestronaux unusable @knro @ckemper67 - the driver seems to be crushing very often (almost immedietely after first sync or slew). I have no data yet - had no time to investigate - but reverting to 1171086 fixed it enough to work for a few hours with a new math plugin (with 12 sync points worked very well few arcmin pointing accuracy and 30s exposures at 1280mm 3.7um pixel essentially 90% perfect). Please take a look what may be a reason. I will add additional info as soon as I have some time to investigate. Note that this may be not limited to caux driver!

knro:
@jochym Any backtrace?

jochym:
Sorry, I know I should post it - but I was in a hurry to make it working first...
I will compile it again on master and provide the logs. On the other hand we are getting better on the CAUX driver. My stacks are really good. The only complain would be less-then perfect pointings (few minutes, but with 10-12 sync points it is getting closer to 1arcmin). BTW - while we are at CAUX - please take a look at https://github.com/indilib/indi-3rdparty/pull/1318 it is a small change which fixes a bug which could be a very confusing surprise for a novice user and paint the driver in a much worse light that it deserves.

ckemper67:
@jochym did you upgrade the indi build as well? #1320 has a dependency on indi#2387 and indi#2388 where I introduced the new methods.

I just realized that I broke ABI compatibility for the math plugins when I introduced the new methods. Do we need to change the version number, so we don't link against an outdated version of the ABI? I think the version skew could cause crashes like this.

jochym:
I think I did, but it was at night and I was trying to make it work ASAP. I will do a full investigation soon.

jochym:
Ok. The instabilities did not return. I think it was a false alarm. If anything more comes up I will open an issue.

Issue #1241, opened 2026-01-08
indi_qsi_ccd : crash after image acquisition end


Jack-adi:
Describe the bug
When acquiring an image using QSI camera, the acquisition start and the driver crash after exposure when getting the image

To Reproduce
Problem happen each time the I try to acquire an image

Desktop (please complete the following information):

    OS: Linux MINT
    Kstars, indilib and 3rd part are compiled from source with debug info (see gdb output)

Log Files
See gdp output attached[
crash_20251205_215349.txt](https://github.com/user-attachments/files/24498698/crash_20251205_215349.txt)

knro:
The logs show buffer overflow detected. So it's a memory issue. Maybe memory allocation issue. What's your camera resolution and image type? Mono or bayered?

Issue #1240, opened 2026-01-08
indi_qsi_ccd : crash after image exposure end


Jack-adi:
**Describe the bug**
When acquiring an image using QSI camera, the acquisition start and the driver crash after  exposure when getting the image

**To Reproduce**
Problem happen each time the I try to acquire an image

**Desktop (please complete the following information):**
 - OS: Linux MINT
 - Kstars, indilib and 3rd part are compiled from source with debug info (see gdb output)

**Log Files**
See gdp output attached

[crash_20251205_215349.txt](https://github.com/user-attachments/files/24498599/crash_20251205_215349.txt)

PR #1329, opened 2026-05-30
Update libqsi version number in debian meta

cgobat:
The QSI SDK provides `get_ImageArraySize()` to query the image array dimensions and element size used for `get_ImageArray()`. The driver previously declared the frame buffer pointer before querying this size and relied on the existing frame buffer allocation, which could be based on stale and/or predicted dimensions rather than the actual size.

This PR revises the `grabImage()` function to call `get_ImageArraySize()` immediately before image download, resize the INDI frame buffer from the SDK-reported `x * y * z` size, and then refresh the frame buffer pointer before passing it to `get_ImageArray()`. This should hopefully avoid the buffer overflow issue reported in #1240, #1241, and indilib/indi#2207.

### DIFF
--- a/debian/libqsi/changelog
+++ b/debian/libqsi/changelog
@@ -1,3 +1,9 @@
+libqsi (7.6.2) noble; urgency=medium
+
+  * New release
+
+ -- Jasem Mutlaq <mutlaqja@ikarustech.com>  Sat, 30 May 2026 20:18:11 +0300
+
 libqsi (7.6.1) bionic; urgency=medium
 
   * New release
--- a/debian/libqsi/libqsi7.install
+++ b/debian/libqsi/libqsi7.install
@@ -1,4 +1,4 @@
-usr/lib/*/libqsiapi.so.7.6.1
+usr/lib/*/libqsiapi.so.7.6.2
 usr/lib/*/libqsiapi.so.7
 usr/bin/qsiapitest
 usr/bin/qsiapidemo


knro:
Interesting, why would this X, Y, or Z be different from what was previously allocated? How come this only affect QSI 600 series cameras? I cannot reproduce any crashes with 500 series and even 600 series (if compiled from source). The crash happens inside the SDK and yesterday I attempted a [fix](https://github.com/indilib/indi-3rdparty/commit/771e2a6fd696191baf022b92d9e6930c9676f645) for it in libqsi directly. Can you check if that works? If not, why would X, Y, or Z be different from previous allocation?

cgobat:
I confirmed that this branch (which includes both this change as well as your recent libqsi change) compiled from source fixes the buffer overflow I have been encountering on a Linux Mint x86 system (w/ a QSI 683). I will try a version that includes only your libqsi change and not this one to try to see which actually resolved it.

knro:
Thank you. Awaiting your confirmation.

cgobat:
Hi Jasem, I tested things with just your libqsi update and it appears to have fixed the issue on its own. To be honest, I did not notice/see that you had just done that when I first made this change/PR, and this update was kind of a shot in the dark that I then ran with because it seemed to have fixed the issue. But I guess it was actually your libqsi change that fixed it, and this update simply didn't break anything further.

I guess that means the buffer overflow issue we have been seeing was probably actually in the diagnostic/log message output rather than the image data itself, so your changes to the `snprintf` buffer/offset logic were what fixed it. I rebased this PR to now only contain a quick update to make the libqsi version in the Debian metadata files match what's in your latest update. My previous commit that modified indi-qsi is apparently not actually needed and has been removed from this changeset.

knro:
Thank you for the confirmation @cgobat . Glad to hear this long elusive crash is finally fixed! Just in time for tomorrow's INDI release.

PR #1338, opened 2026-06-15
Improves/updates/fixes indi-gige driver

olsonse:
Most of the code in the indi-gige driver was written specifically for at least one older version of a point-grey BlackFly camera using an older version of the Aravis library.  These patches update the indi-gige driver to use the more recent and common 0.8 version of the Aravis library.  These patches also greatly improve the driver to be more generically applicable to other cameras.  These patches also allow for streaming from an Aravis-supported camera (either GigE-Vision or USB-vision).  These patches provide the necessary changes that will also allow for multiple cameras to be used simultaneously.  Nevertheless, these patches do not implement the iterative enumeration that will be necessary to instantiate all available cameras (only the first is still instantiated)--it is left as a task for a future patch to implement this extended instantiation.

### DIFF
--- a/indi-gige/AUTHORS
+++ b/indi-gige/AUTHORS
@@ -1,1 +1,2 @@
 Hendrik Beijeman  (hbeyeman@gmail.com) http://phym.nl
+Spencer E. Olson (olsonse@umich.edu)
--- a/indi-gige/README
+++ b/indi-gige/README
@@ -15,7 +15,7 @@ Requirements
 
 	cfitsio-devel is required to compile support for FITS.
 	
-+ aravis >= v0.6
++ aravis >= v0.8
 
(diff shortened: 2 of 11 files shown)


knro:
Thank you! Can you please submit driver documentation to [INDI Documentation Repo](https://github.com/indilib/drivers-docs)?

olsonse:
Yes, I think I can add something here.  First question is, looking through the [INDI Driver Docs Repo](https://github.com/indilib/drivers-docs), it seems that GiGE and USB3-Vision cameras will best be suited in directory for the "generic" cameras.  Any contention with this?

knro:
No issues at all.

olsonse:
I just submitted a pull request for the driver documentation in the [INDI Driver Docs Repo](https://github.com/indilib/drivers-docs).

PR #1340, opened 2026-06-21
gphoto: fix range type manual focusing

MmAaXx500:
- `gp_widget_set_value` expects float value for RANGE type widgets
 - reset the value to 0 before performing the focusing (similar way to RADIO type widget) otherwise the action is not performed by the camera if the same value is used repeatedly

Tested with a Sony a6400


knro:
Thank you! Let me see if this works for other cameras as well.

MmAaXx500:
Crosslinking kstars report:
EKOS: Relative focus adjustment for Sony cameras uses wrong Indi command https://bugs.kde.org/show_bug.cgi?id=516680

PR #1343, opened 2026-06-26
Add  target liveviewFPS parameters

Nekomeshi:
In the current implementation, the next frame acquisition is triggered immediately after a frame is received from the camera. As a result, when using older cameras such as the Nikon D5300, the camera cannot keep up with the continuous acquisition requests, causing it to freeze.

To address this issue, a new parameter will be added to acquire frames at a specified frame-rate interval. In addition, another parameter will be introduced to insert a wait period after each frame acquisition, allowing extra time for the camera to complete its internal processing when the specified frame-rate interval alone is not sufficient.

### DIFF
--- a/indi-gphoto/gphoto_ccd.cpp
+++ b/indi-gphoto/gphoto_ccd.cpp
@@ -43,6 +43,7 @@
 #define MAX_DEVICES  5 /* Max device cameraCount */
 #define FOCUS_TIMER  50
 #define MAX_RETRIES  3
+static const char * STREAM_TAB = "Streaming"; //Original is declared in streammanager.cpp
 
 #ifdef __APPLE__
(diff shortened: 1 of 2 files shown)


knro:
Right, I noticed that my Nikon D3500 generates like 160fps which is too high. However, no need to introduce new properties. There is already STREAMING_EXPOSURE and STREAM_DELAY defined in INDI::StreamManager so these value can be used to throttle down very high FPS from such cameras.

Nekomeshi:
Thank you for the advice. STREAMING_EXPOSURE is usable for setting FPS. But STREAM_DELAY  is a read-only property.

Nekomeshi:
I modified to use  STREAMING_EXPOSURE property for setting target FPS. As STREAM_DELAY is read only property, I'm still using dedicated property for setting the delay, but changed the property name.

PR #1345, opened 2026-06-30
indi-libcamera:  Add Gain Conversion control for IMX290-family sensors

OLEGsandrMB:
## Summary

This PR adds a **Gain Conversion** property to the `indi-libcamera` driver for Sony IMX290-family image sensors.

The property allows switching between:

* **Dynamic Range (LCG)**
* **Low Noise (HCG)**

The driver reads the current gain conversion mode when the camera connects, updates the INDI property accordingly, and writes the selected mode through the existing kernel interface.

## Supported sensors

This feature is intended for Sony IMX290-family sensors, such as:

* IMX290
* IMX327
* IMX462

## Motivation

These sensors support two analog gain conversion modes.

* **LCG (Low Conversion Gain)** provides higher dynamic range.
* **HCG (High Conversion Gain)** provides lower read noise, which is especially useful for astrophotography.

Until now this functionality was not available through the INDI LibCamera driver.

## Notes

The current implementation uses the existing kernel parameter:

`/sys/module/imx290/parameters/hcg_mode`

Writing this parameter requires appropriate system permissions. If writing fails, the driver reports an informative error message.

A future improvement would be exposing Gain Conversion as a standard V4L2/libcamera control, removing the need for direct sysfs access.

## Tested

Tested successfully on:

* Raspberry Pi
* Sony IMX462
* Raspberry Pi `imx290` kernel driver
* INDI Control Panel

Verified:

* Current mode is read correctly on connection.
* Switching between Dynamic Range (LCG) and Low Noise (HCG) updates the kernel parameter.
* The INDI property stays synchronized with the actual hardware state.

### DIFF
--- a/indi-libcamera/README.md
+++ b/indi-libcamera/README.md
@@ -28,6 +28,34 @@ The Driver can run multiple devices if required, to run the driver:
 
 AVAILABLE CONTROLS
 
+## Gain Conversion (HCG/LCG)
+
+The Gain Conversion control is available for IMX290-family sensors when the
+kernel exposes the following sysfs interface:
+
+    /sys/module/imx290/parameters/hcg_mode
+
+When this interface is available, indi-libcamera exposes a "Gain Conversion"
+property that allows switching between:
+
+- Dynamic Range (LCG)
+- Low Noise (HCG)
+
+If the sysfs interface is not present, the property is hidden.
+
+By default, the sysfs interface is typically writable only by root.
+
+One way to allow a non-root `indiserver` to control Gain Conversion is to
+change the group ownership to a group that the `indiserver` user belongs to,
+for example:
+
+```bash
+sudo chgrp video /sys/module/imx290/parameters/hcg_mode
+```
+
+For a permanent configuration, use a udev rule or another system-specific
+mechanism to assign the desired group automatically.
+
 TODO 
 
 You can also start video stream.
--- a/indi-libcamera/indi_libcamera.cpp
+++ b/indi-libcamera/indi_libcamera.cpp
@@ -43,8 +43,17 @@
 #include <libraw.h>
 #include <jpeglib.h>
 
+#include <fstream>
+#include <cerrno>
+#include <cstring>
 
 #define CONTROL_TAB "Controls"
+
+namespace
+{
+constexpr const char *HCG_SYSFS =
+    "/sys/module/imx290/parameters/hcg_mode";
+}
 
 static class Loader
 {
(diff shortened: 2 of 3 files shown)


knro:
Thank you, shouldn't the property be defined only if ```/sys/module/imx290/parameters/hcg_mode``` exists?

OLEGsandrMB:
Thanks for the suggestion. I've updated the driver so the Gain Conversion property is only defined when readGainConversionMode() succeeds, i.e. when the hcg_mode sysfs interface is available. When the interface is absent, the property is not exposed.

geoffrey-vl:
Thanks, I've been thinking of adding it myself but lacked the time, nice that someone picked it up and good to known that there are more users of this camera family :)

geoffrey-vl:
Now to come back to this, how did you work around the sudo permissions needed to write this sysfs interface?

```
$ ls -al /sys/module/imx290/parameters/hcg_mode
-rw-rw-r-- 1 root root 4096 Jul 11 23:57 /sys/module/imx290/parameters/hcg_mode
```

OLEGsandrMB:
Thanks for pointing that out. During development I was running indiserver as root, so I hadn't considered the permissions issue for normal deployments. I agree that relying on a writable sysfs parameter isn't ideal. Would you prefer this to be exposed as a V4L2 control instead?

geoffrey-vl:
I think we both know that this would be the most ideal solution, but than again I believe the [following comment on the PRI forums](https://forums.raspberrypi.com/viewtopic.php?p=2182756&hilit=exposure#p2186823) explains pretty much that the succes rate for that is rather low. Feel free to do, meanwhile I think we could be pragmatic and onboard the current solution that you made.

I had a look in the code for usage of sudo and probable the closest thing I could find is [indi_gpio.cpp](https://github.com/indilib/indi-3rdparty/blob/ef6f518e50bd9a1a926a912fb39d3faf5c4871a6/indi-gpio/indi_gpio.cpp#L979). I think if you make a similar error msg it would be OK. Additionally I think an extra section in the README.md for indi-libcamera could make some sense here to explain how the HCG mode can be used and what requirements it has. Some lines of example commands that the user should do to setup correct permissions would be nice I think.

OLEGsandrMB:
Thanks for the review!

I've updated the PR based on your suggestions:

The Gain Conversion property is now defined only when the hcg_mode sysfs interface is available.
Improved the error message when the sysfs interface is not writable.
Added a README section describing the feature, its requirements, and an example of configuring access to the sysfs interface.

I've tested the updated implementation on an IMX462-based camera.

indi-libcamera/README.md
```diff
@@ -28,6 +28,34 @@ The Driver can run multiple devices if required, to run the driver:
...
+If the sysfs interface is not present, the property is hidden.
+
+By default, the sysfs interface is typically writable only by root.
+
+One way to allow a non-root `indiserver` to control Gain Conversion is to
+change the group ownership to a group that the `indiserver` user belongs to,
+for example:
+
+```bash
+sudo chgrp video /sys/module/imx290/parameters/hcg_mode
```
geoffrey-vl, line 53: valid: on RaspberryPi OS the pi user  is member of the video group

geoffrey-vl (approved):
LGTM

PR #1347, opened 2026-07-04
Fix eqmod crashes with old 150i MC firmware

### LINKED ISSUE #1346: Constant crashes with eqmod driver and 150i mount
**Describe the bug**
v2.2.3.1 produces a stream of crashes/driver disconnections when the eqmod driver is used with a Skywatcher 150i. At first I had the impression, that the latest modification in 3rdparty did this, but unfortunately even after reverting it the symptoms persists.
This was present on both 3.58 and 3.66 firmware versions of the mount.
Some times it is a crash, but most of the times it is just a lost connection. Full restart of indi/kstars is needed for recovery.

**To Reproduce**
Exact steps to reproduce the behavior.
Start the driver. Connect PHD2, slew to some position, start guiding. within minutes the mount will get disconnected.

**Expected behavior**
Driver runs stable.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Arch linux ARM
 - Version 2.2.3.1

**Log Files**

gy-lehel:
eqmod driver with a SW 150i running 3.58 MC firmware was driven into a continuous crash loop. This fix should omit the indexer check with this old firmware revision.

Fixes: #1346

### DIFF
--- a/indi-eqmod/skywatcher.cpp
+++ b/indi-eqmod/skywatcher.cpp
@@ -210,7 +210,14 @@ uint32_t Skywatcher::GetDEPeriod()
 uint32_t Skywatcher::GetlastreadRAIndexer()
 {
     if (MountCode != 0x04 && MountCode != 0x05 && MountCode != 0x20 && MountCode != 0x25 && MountCode != 0x45)
+    {
         throw EQModError(EQModError::ErrInvalidCmd, "Incorrect mount type");
+    }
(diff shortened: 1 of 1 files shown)


gy-lehel:
I'd like to ask someone with a better knowledge of the MCVersion attribute to double-crosscheck if my assumption of the version number is correct. I took it from an old issue, where I reported the number seen in the ekos GUI.

knro:
How old is 3.58?

PR #1350, opened 2026-07-19
indi-playerone: apply STREAMING_EXPOSURE updates mid-stream

dorothy2846:
### Problem

`POABase::workerStreamVideo` reads `Streamer->getTargetFPS()` once, before the capture loop starts. Any `STREAMING_EXPOSURE_VALUE` a client writes **while streaming** updates the StreamManager's target FPS but never reaches the camera — the worker keeps exposing at the value captured at stream start. The only way to change streaming exposure is to stop and restart the stream, which is disruptive for live view / EAA use ("adjust brightness" becomes "drop the stream").

### Fix

Re-poll `Streamer->getTargetFPS()` at the top of each loop iteration. When it changes, push the new exposure to the camera (`POA_EXP` / `POA_EXPOSURE`, matching the existing 0.95 duty factor) and update the frame-wait timeout accordingly.

### Testing

Player One Uranus-C on a headless INDI setup: started streaming, wrote `STREAMING_EXPOSURE_VALUE` repeatedly mid-stream from the client. Before: image brightness never changed until the stream was restarted. After: exposure follows each update within a frame, stream uninterrupted. Also verified normal still exposures and stream start/stop are unaffected.

### DIFF
--- a/indi-playerone/playerone_base.cpp
+++ b/indi-playerone/playerone_base.cpp
@@ -149,7 +149,8 @@ bool POABase::stopExposure()
 void POABase::workerStreamVideo(const std::atomic_bool &isAbortToQuit)
 {
     POAErrors ret;
-    double ExposureRequest = 1.0 / Streamer->getTargetFPS();
+    double currentTargetFPS = Streamer->getTargetFPS();
+    double ExposureRequest = 1.0 / currentTargetFPS;
     POAConfigValue confVal;
 #ifdef USE_POA_EXP
     confVal.floatValue = static_cast<double>(ExposureRequest * 0.95);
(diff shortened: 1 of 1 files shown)


knro:
I'm not sure how efficient this approach is. Maybe using a signal is more efficient.

hiro3110i:
@dorothy2846 Thank you for your suggestion.
I confirmed your idea to PlayerOne engineer whether we can change exposure during video streaming.
We can change exposure time, gain and offset after starting exposure. Binning and ROI can not change due to different image size.

I'd like to implement your idea simpler with appropriate error handling if we can get the latest FPS in exposure loop.
Is it acceptable?

@knro I'd like to confirm return value of getTargetFPS() could be changed during streaming.

hiro3110i:
@dorothy2846 
I sent PR including exposer change during video streaming.
Please look at https://github.com/indilib/indi-3rdparty/pull/1370
Refactoring is the main aim of this PR, but including you suggestion.

cfuture81:
I guess this one can be closed now as @hiro3110i 's PR is now merged.

https://github.com/indilib/indi-3rdparty/pull/1370

PR #1351, opened 2026-07-20
Fix CP210x DTR/RTS reset on ESP32 (HBG3 adapter) in indi-celestronaux

jochym:
## Summary
Fixes periodic microcontroller reset issue when using Celestron AUX driver with Silicon Labs CP210x USB-UART bridges (e.g., HBG3 adapter) connected to ESP32-based devices.

## Root Cause
The CP210x kernel driver forces DTR/RTS lines LOW (active) on port open. On hardware with auto-reset circuits (DTR→EN, RTS→GPIO0), this causes a hard ESP32 reset on every connection/reconnection.

## Solution
Added  method that sets termios flags immediately after port open:
-  - prevents DTR/RTS assertion on open
-  - prevents DTR/RTS drop on close  
- Conditional  - enabled only for AUX/PC port (19200 baud, RTS/CTS half-duplex), disabled for HC/USB port (9600 baud)

Flags preserved in  during baud rate changes.

## Changes
- : +1 line (method declaration)
- : +73 lines (implementation + Handshake/tty_set_speed integration)

## Testing
- AUX/PC port (19200 baud): RTS/CTS half-duplex preserved
- HC/USB port (9600 baud): No flow control, no reset
- Both modes verified working on HBG3 adapter hardware and HC/serial connection.

### DIFF
--- a/indi-celestronaux/CHANGELOG.md
+++ b/indi-celestronaux/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog - indi-celestronaux System Tests
 
+## [2026-07-20 20:15] - Version 1.7.0 - CP210x DTR/RTS Reset Fix
+- **Fixed CP210x microcontroller reset issue** on ESP32-based mounts (HBG3 adapter)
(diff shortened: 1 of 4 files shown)


jochym:
Note that the failing build is not connected with this driver.

PR #1352, opened 2026-07-21
Add CAA Rotator support

278261631:
rotate the camera, negative angles for reverse rotation


knro:
Thank you. I didn't realize there are camera with built-in rotators from QHY.

PR #1354, opened 2026-07-22
indi-asi: Implement gradual warm-up when turning the cooler off

pmneo:
Previously, switching the Cooler OFF immediately cut TEC power and set CoolerSP to IPS_IDLE, bypassing the temperature-ramp machinery entirely. This could cause thermal shock on sensors that were significantly below ambient temperature.

This commit introduces a full warm-up state machine that mirrors the existing cooling ramp (driven by TemperatureRampNP RAMP_SLOPE / RAMP_THRESHOLD in the INDI::CCD base class) and keeps the Cooler switch in IPS_BUSY until the camera has safely returned to near-ambient temperature.

New behaviour
─────────────
• Cooler OFF with RAMP_SLOPE = 0 (ramping disabled): unchanged, TEC
  cuts immediately.
• Cooler OFF with RAMP_SLOPE > 0: begins a gradual warm-up toward the
  configurable "Warm Up Target" temperature (CCD_WARMUP_TARGET, default
  25 °C, persisted in the driver config).  CoolerSP stays IPS_BUSY
  (switch shows OFF) until complete.
• Warm-up completion is detected by two independent criteria, whichever
  fires first:
    1. Temperature reading reaches within RAMP_THRESHOLD of the warm-up target — handled by the existing INDI::CCD::checkTemperatureTarget() override.
    2. Temperature has been stable (changed less than 0.5 °C) for 60 s — catches the case where the actual room temperature is below the configured warm-up target, so the camera equilibrates below 25 °C and would otherwise never cross the threshold. • Cooler ON during an active warm-up: cancels the warm-up immediately
  and resumes cooling toward the previously set target temperature,
  respecting the ramp slope.
• An explicit temperature request (TemperatureNP) while warming up also
  cancels the warm-up and hands control back to the standard ramp path.

Key implementation details
──────────────────────────
• Split activateCooler() into setCoolerPower() (raw TEC toggle, no UI
  update) and activateCooler() (TEC toggle + CoolerSP state update).
  SetTemperature() calls setCoolerPower() during warm-up so that the
  ~60 s ramp steps driven by m_TemperatureCheckTimer do not flip
  CoolerSP back to ISS_ON/IPS_BUSY on every iteration.
• mSavedCoolingTarget captures m_TargetTemperature (the base-class
  member set by ISNewNumber, holding the user's final cooling setpoint)
  at the start of beginCoolerWarmup(), before it is overwritten with the
  warm-up target.  mTargetTemperature (ASIBase's own member) was
  intentionally not used here because it only tracks the last ramp step,
  not the user's original request.
• ASI TECs are unidirectional (cooling only); TEC power reads 0 % for
  the entire warm-up ramp, so a raw "power idle" check is not a useful
  completion signal.  The stability-based secondary check avoids this
  pitfall.

New property
────────────
CCD_WARMUP_TARGET ("Warm Up Target") — configurable warm-up target temperature in °C, shown in the main control tab for cooled cameras. Saved to and restored from the driver configuration file.

### DIFF
--- a/indi-asi/asi_base.cpp
+++ b/indi-asi/asi_base.cpp
@@ -41,6 +41,9 @@
 #define VERBOSE_EXPOSURE        3
 #define TEMP_TIMER_MS           1000 /* Temperature polling time (ms) */
 #define TEMP_THRESHOLD          .25  /* Differential temperature threshold (C)*/
+#define TEMP_AMBIENT_DEFAULT    25.0  /* Default warm-up target on cooler off (C), ASI SDK has no ambient sensor */
+#define WARMUP_STABLE_MS        60000 /* Temperature must be stable this long before warm-up is considered done */
+#define WARMUP_STABLE_DELTA     0.5   /* Maximum temperature change (C) that counts as "stable" */
 
 #define CONTROL_TAB "Controls"
 
@@ -403,6 +406,9 @@ bool ASIBase::initProperties()
     CoolerNP[0].fill("CCD_COOLER_VALUE", "Cooling Power (%)", "%+06.2f", 0., 1., .2, 0.0);
     CoolerNP.fill(getDeviceName(), "CCD_COOLER_POWER", "Cooling Power", MAIN_CONTROL_TAB, IP_RO, 60, IPS_IDLE);
 
+    WarmupTargetNP[0].fill("TEMPERATURE", "Temperature (C)", "%.1f", -50., 50., 1., TEMP_AMBIENT_DEFAULT);
+    WarmupTargetNP.fill(getDeviceName(), "CCD_WARMUP_TARGET", "Warm Up Target", MAIN_CONTROL_TAB, IP_RW, 60, IPS_IDLE);
+
     ControlNP.fill(getDeviceName(), "CCD_CONTROLS",      "Controls", CONTROL_TAB, IP_RW, 60, IPS_IDLE);
     ControlSP.fill(getDeviceName(), "CCD_CONTROLS_MODE", "Set Auto", CONTROL_TAB, IP_RW, ISR_NOFMANY, 60, IPS_IDLE);
 
@@ -505,6 +511,8 @@ bool ASIBase::updateProperties()
             loadConfig(true, CoolerNP.getName());
             defineProperty(CoolerSP);
             loadConfig(true, CoolerSP.getName());
+            defineProperty(WarmupTargetNP);
+            loadConfig(true, WarmupTargetNP.getName());
         }
         // Even if there is no cooler, we define temperature property as READ ONLY
         else
@@ -566,6 +574,7 @@ bool ASIBase::updateProperties()
         {
             deleteProperty(CoolerNP);
             deleteProperty(CoolerSP);
+            deleteProperty(WarmupTargetNP);
         }
         else
             deleteProperty(TemperatureNP);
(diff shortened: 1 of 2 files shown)


knro:
What's wrong with using the existing system as-is?

pmneo:
It is not warming ;) OFF just turns it off without using the ramping settings. It would only work if the temp is set to 25 or something ...

pmneo:
My ASI 2600 daughterbord died twice within one year now and the suggestion of ZWO is to warm the camera in slow steps instead of turning it off ...

knro:
Right, but if you set the temperature ramp to say 5C in the driver, then if it was -20 and target is 10C, it would warm it up 5C every minute. This already works as-is.

pmneo:
Yes, but KStars does not use this ;) and the cooler itself is never disabled

pmneo:
My fix is applying this logic on the driver level to protected the camera to get distroyed. This missing feature cost me about 1000€ meanwhile :-/

pmneo:
A full roundtrip look now like this:

```
2026-07-22T11:11:57: [INFO] Cooler is now off, temperature has stabilized. 
2026-07-22T11:11:44: [INFO] Setting temperature to 25.00 C. 
2026-07-22T11:10:39: [INFO] Setting temperature to 25.00 C. 
2026-07-22T11:09:39: [INFO] Setting temperature to 25.00 C. 
2026-07-22T11:08:34: [INFO] Setting temperature to 25.00 C. 
2026-07-22T11:07:29: [INFO] Setting temperature to 23.00 C. 
2026-07-22T11:06:24: [INFO] Setting temperature to 18.10 C. 
2026-07-22T11:05:24: [INFO] Cooler is warming up to ambient temperature (25.0 C)... 
2026-07-22T11:05:24: [INFO] Setting temperature to 15.30 C. 
-> turned cooler "off"
2026-07-22T11:02:01: [INFO] Setting temperature to 10.00 C. 
2026-07-22T11:01:01: [INFO] Setting temperature to 10.00 C. 
2026-07-22T11:00:01: [INFO] Setting temperature to 13.30 C. 
2026-07-22T10:59:01: [INFO] Setting temperature to 18.60 C. 
-> cool to 10deg
``

knro:
A simpler solution would be to turn off cooling on warm up once target is reached.

pmneo:
But you don't know the ambient temperature as the ASI Camera has no second sensor for this ...

pmneo:
@knro any other thoughts?

knro:
Right, but you can set the warm up target to 15C for example, and once the target is reached, then we turn off the cooler.

Now the only issue is that if the ambient temperature happens to be lower (e.g. 10C) then the camera cannot warm up to 15C, so we need to implement a timer that would fire off and conclude the operation after a set timeout.

pmneo:
> Right, but you can set the warm up target to 15C for example, and once the target is reached, then we turn off the cooler.
> 
> Now the only issue is that if the ambient temperature happens to be lower (e.g. 10C) then the camera cannot warm up to 15C, so we need to implement a timer that would fire off and conclude the operation after a set timeout.

But this is exactly what I have implemented here :) We try to warm up to 25 deg (configable). 
If the temp does then not warm more the timer steps in and `2026-07-22T11:11:57: [INFO] Cooler is now off, temperature has stabilized.`  is logged and the cooler is turned of in real.

knro:
Right, but there is no need to introduce a new property for this. Furthermore, such a solution should be in core INDI and not a single driver, so that all camera drivers benefit from this and behave the same. I'll make some changes in INDI::CCD class. Thank you for the suggestion!

pmneo:
Closed means, you will implement this in the core? My analysis showed, that there is no real common basis in indilib for doing this driver independent?

PR #1359, opened 2026-07-22
Fix for A7RIV timeout & capturing a single frame only

cbreeves:
When attempting to use the Sony A7RIV, I encountered a timeout failure which caused an extensive delay and eventual timeout, with the result that of any 1..N frames, frames 2..N failed to capture.   

I traced it to this ptp session reset between each frame.   The change attached to this PR prevents this session reset for ILCE-7RM4, the only A7R model for which I can confirm the reset is destructive to sequence captures.   However, I will say that to me at least, if feels highly unlikely that the reset operation is useful on any model, given how long it takes to complete.

This is the driver code I have been running for my last 4-5 sessions, and it is working reliably for hundreds of frames in sequence.


knro:
This was explicitly added for these models, so not sure why your model is different. At any rate, actual testing trumps any theories, so I'll merge this, thanks!

PR #1360, opened 2026-07-23
indi-asi: Implement gradual warm-up when turning the cooler off

pmneo:
Previously, switching the Cooler OFF immediately cut TEC power and set
CoolerSP to IPS_IDLE, bypassing the temperature-ramp machinery entirely.
This could cause thermal shock on sensors that were significantly below
ambient temperature.

This commit introduces a full warm-up state machine that mirrors the
existing cooling ramp (driven by TemperatureRampNP RAMP_SLOPE /
RAMP_THRESHOLD in the INDI::CCD base class) and keeps the Cooler switch
in IPS_BUSY until the camera has safely returned to near-ambient
temperature.

New behaviour
─────────────
• Cooler OFF with RAMP_SLOPE = 0 (ramping disabled): unchanged, TEC
  cuts immediately.
• Cooler OFF with RAMP_SLOPE > 0: begins a gradual warm-up toward the
  configurable "Warm Up Target" temperature (CCD_WARMUP_TARGET, default
  25 °C, persisted in the driver config).  CoolerSP stays IPS_BUSY
  (switch shows OFF) until complete.
• Warm-up completion is detected by two independent criteria, whichever
  fires first:
    1. Temperature reading reaches within RAMP_THRESHOLD of the warm-up
       target — handled by the existing INDI::CCD::checkTemperatureTarget()
       override.
    2. Temperature has been stable (changed less than 0.5 °C) for 60 s —
       catches the case where the actual room temperature is below the
       configured warm-up target, so the camera equilibrates below 25 °C
       and would otherwise never cross the threshold.
• Cooler ON during an active warm-up: cancels the warm-up immediately
  and resumes cooling toward the previously set target temperature,
  respecting the ramp slope.
• An explicit temperature request (TemperatureNP) while warming up also
  cancels the warm-up and hands control back to the standard ramp path.

Key implementation details
──────────────────────────
• Split activateCooler() into setCoolerPower() (raw TEC toggle, no UI
  update) and activateCooler() (TEC toggle + CoolerSP state update).
  SetTemperature() calls setCoolerPower() during warm-up so that the
  ~60 s ramp steps driven by m_TemperatureCheckTimer do not flip
  CoolerSP back to ISS_ON/IPS_BUSY on every iteration.
• mSavedCoolingTarget captures m_TargetTemperature (the base-class
  member set by ISNewNumber, holding the user's final cooling setpoint)
  at the start of beginCoolerWarmup(), before it is overwritten with the
  warm-up target.  mTargetTemperature (ASIBase's own member) was
  intentionally not used here because it only tracks the last ramp step,
  not the user's original request.
• ASI TECs are unidirectional (cooling only); TEC power reads 0 % for
  the entire warm-up ramp, so a raw "power idle" check is not a useful
  completion signal.  The stability-based secondary check avoids this
  pitfall.

New property
────────────
CCD_WARMUP_TARGET ("Warm Up Target") — configurable warm-up target
temperature in °C, shown in the main control tab for cooled cameras.
Saved to and restored from the driver configuration file.

### DIFF
--- a/indi-apogee/apogee_ccd.cpp
+++ b/indi-apogee/apogee_ccd.cpp
@@ -322,13 +322,13 @@ bool ApogeeCCD::getCameraParams()
     return true;
 }
 
-int ApogeeCCD::SetTemperature(double temperature)
+int ApogeeCCD::SetTemperature(double temperature, bool enableCooler)
 {
     // If less than 0.1 of a degree, let's just return OK
     if (fabs(temperature - TemperatureNP[0].getValue()) < 0.1)
         return 1;
 
-    if (!m_CoolerWarmingUp)
+    if (enableCooler)
         activateCooler(true);
 
     try
--- a/indi-apogee/apogee_ccd.cpp
+++ b/indi-apogee/apogee_ccd.cpp
@@ -328,7 +328,8 @@ int ApogeeCCD::SetTemperature(double temperature)
     if (fabs(temperature - TemperatureNP[0].getValue()) < 0.1)
         return 1;
 
-    activateCooler(true);
+    if (!m_CoolerWarmingUp)
+        activateCooler(true);
 
     try
     {
@@ -437,9 +438,17 @@ bool ApogeeCCD::ISNewSwitch(const char *dev, const char *name, ISState *states, 
                 return false;
 
             if (CoolerS[0].s == ISS_ON)
+            {
+                cancelCoolerWarmup();
                 activateCooler(true);
+                resumeCoolingAfterWarmup();
+            }
             else
-                activateCooler(false);
+            {
+                CoolerSP.s = IPS_BUSY;
+                beginCoolerWarmup(TemperatureNP[0].getValue());
+                IDSetSwitch(&CoolerSP, nullptr);
+            }
 
             return true;
         }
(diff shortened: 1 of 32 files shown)


pmneo:
@knro Based on https://github.com/indilib/indi/pull/2453

pmneo:
@knro any thougths on this? I put a high effort into this to implement it globaly as you mentioned?

knro:
Thanks, I already implemented the solution in indi-core, did you check it out?

pmneo:
No ... have you seen my PR https://github.com/indilib/indi/pull/2453/ ? I have implemented gradual warming if ramping is enabled when turning the cooler to "off" and tried to implement it on all drivers.

pmneo:
I have successfully runned this both branches on my rig since the PR

knro:
The solution implemented now only affects indi-core and does not require changes to INDI 3rd party driver.

pmneo:
But it does not work with the cooler-off switch, right?

knro:
It doesn't explicitly set cooler off, but if cooler is at 0% , that's almost the same thing. The cooler off change required substantial changes which would not be suitable as the next INDI release is just around the corner. We will revisit this for the next INDI development cycle after Aug 1st release.

PR #1363, opened 2026-08-08
indi-asi: make ZWO CAA max-limit read fail-safe

cfuture81:
## Summary

Make the ZWO CAA rotator's max-degree limit read fail-safe.

`CAAGetMaxDegree()` can fail transiently (e.g. a USB glitch on connect). Previously a failed read left the internal limit at `0`, after which `MoveRotator()` rejected **every** subsequent move with `Target angle ... exceeds max limit 0.00` — effectively bricking motion until reconnect.

## Changes

- `Connect()`: on a failed or non-positive `CAAGetMaxDegree()`, fall back to `360` (with a warning) instead of leaving the limit at `0`.
- `MoveRotator()`: only enforce a **positive** limit, so an unset/unreadable limit can no longer reject all motion.

## Notes

Small, self-contained robustness fix, independent of any other change. Builds and links against libindi. The behaviour change is limited to the failure path; nominal operation (valid limit read) is unchanged.

### DIFF
--- a/indi-asi/asi_rotator.cpp
+++ b/indi-asi/asi_rotator.cpp
@@ -257,16 +257,25 @@ bool ASICAA::Connect()
     else
         ReverseRotatorSP.setState(IPS_ALERT);
 
-    // Get max degree limit
+    // Get max degree limit. If the read fails (or returns a non-positive,
+    // unusable value), fall back to a full turn. Otherwise a transient read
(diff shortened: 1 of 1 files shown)


knro:
Thanks, was this tested on actual hardware?

cfuture81:
> Thanks, was this tested on actual hardware?

User test pending (discord thread with Sparky)

knro:
Great, let me know once it's confirmed.

PR #1364, opened 2026-08-08
indi-asi: migrate ZWO CAA rotator to HotPlugManager framework

cfuture81:
## Summary

Migrate the ZWO CAA rotator driver to the shared `HotPlugManager` / `HotPlugCapableDevice` framework already used by the ASI CCD, focuser (EAF) and filter wheel (EFW) drivers.

## Motivation

The CAA rotator is the **only** ASI driver still using a bespoke static `Loader`. That loader enumerates devices once at startup, and its `hotPlugTimer` is declared but never wired — so the driver does **no in-process hotplug detection**. The other ASI drivers register with `HotPlugManager` (periodic poll, create/destroy device objects live), so they handle unplug/replug while the driver is running. The rotator does not: a replugged rotator is not re-detected until the driver process is restarted, which is inconsistent with the rest of the ASI family.

## Changes

- Add `ASICAAHotPlugHandler` (`HotPlugCapableDevice`) implementing `discoverConnectedDeviceIdentifiers` / `createDevice` / `destroyDevice` / `getManagedDevices` via the CAA SDK. Discovery and creation deliberately avoid `CAAOpen()` so enumeration never claims the USB interface.
- Replace the custom `Loader` with the standard `registerHandler()` / `start()` wiring.
- Give `ASICAA` a `(CAA_INFO, name, serial)` constructor with `getCAAInfo()` / `getSerialNumber()` accessors, matching the other ASI device classes. The historical `"ZWO CAA <Name>"` device-name scheme is preserved so existing saved configurations keep working.
- Add the new source to the `indi_asi_rotator` CMake target.

## Scope / notes

- This is a **consistency / modernization** change — it brings the rotator's hotplug behaviour in line with the other ASI drivers. It is **not** a fix for any specific field failure.
- Builds and links against libindi. **Not yet validated on hardware** — the meaningful test is an **unplug/replug** of the CAA with the driver running, confirming it re-detects in-process like the camera/focuser/wheel.
- Mirrors `asi_focuser_hotplug_handler.{h,cpp}` closely.

### DIFF
--- a/indi-asi/CMakeLists.txt
+++ b/indi-asi/CMakeLists.txt
@@ -102,7 +102,12 @@ target_link_libraries(indi_asi_focuser ${INDI_LIBRARIES} ${HIDAPILIB} ${ASI_LIBR
 ENDIF()
 
 ########### indi_asi_rotator ###########
-add_executable(indi_asi_rotator ${CMAKE_CURRENT_SOURCE_DIR}/asi_rotator.cpp)
+set(indi_asi_rotator_SRCS
+   ${CMAKE_CURRENT_SOURCE_DIR}/asi_rotator.cpp
+   ${CMAKE_CURRENT_SOURCE_DIR}/asi_rotator_hotplug_handler.cpp
+   )
+
+add_executable(indi_asi_rotator ${indi_asi_rotator_SRCS})
 IF (APPLE)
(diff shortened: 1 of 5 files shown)


knro:
Thank you! Looks good but we weed hardware verification test.

PR #1371, opened 2026-08-13
indi-scopelink: driver for the Astrolabs ScopeLink telescope controller

tothbastrolabs:
## What this is

A driver for the ScopeLink, a single-box telescope controller from Astrolabs. One USB CDC-ACM
connection drives a stepper focuser, a motorised front flap over the aperture, a flat panel, two mirror
cooling fans, two auxiliary power outputs and, on generation 3, a powered USB hub with per-port fault
reporting.

It is published as one INDI device carrying four interfaces — focuser, dust cap, light box and
auxiliary — so Ekos runs autofocus, opens and closes the flap and takes flat frames without further
setup. Which interfaces it claims is decided at connect time from the controller's own identification
block, so a generation 2 unit simply has no USB hub or smart switch properties rather than having ones
that never read.

I am the author and maintainer, and Astrolabs makes the hardware.

## On placement

The README lists three criteria, any one of which puts a driver here, and this driver meets the third:
it is vendor-maintained and wants a release cadence independent of INDI Core.
**ScopeLink has no external dependency.** It builds against libindi alone.

## Layout

- `indi-scopelink/libscopelink/` — the protocol core: framing, retries, identification, capability
  discovery, status, configuration identifiers, fault store and a simulated controller. No INDI, no
  external dependency, so it builds and is tested on its own.
- `indi-scopelink/indi/` — the driver: property definitions and the mapping onto INDI's interfaces.
- `indi-scopelink/tools/` — `scopelink-cli` for support work, and `scopelink-simulator`, which serves
  the simulated controller on a pseudo terminal.
- `indi-scopelink/man/` — manual pages for the driver and both tools. User documentation is
  submitted separately to indilib/drivers-docs#38, as requested in review.
- `debian/indi-scopelink/` — packaging, modelled on `debian/indi-rolloffino`.

The core is kept separate from the driver deliberately: it has to agree byte for byte with a driver
written in another language for another platform, and that agreement is only testable if it can be
exercised without libindi in the way.

## Testing

Against real hardware, both supported generations:

- **A 16 hour unattended soak** on a generation 3 unit at the 500 ms default: 108236 transactions, 0
  retried attempts, 0 failed transactions, connected in all 198 five-minute samples, resident memory
  flat to the kilobyte from first sample to last.
- **Ekos end to end.** The focus module owns the focuser and commanded real moves; a flat frame
  sequence ran the whole path unattended — park dust cap, driver closes the flap, light box on, frames,
  complete.
- **Conformance vectors.** Four recordings captured from real controllers of both generations are
  replayed through the same `Device` the driver uses and checked against expectations worked out by
  hand from the bytes. They cover both status frame layouts (60 and 56 bytes) and both fault store
  freeze frame layouts (48 and 34 bytes, 36 and 24 fields). This is what proves the driver takes each
  layout from the identification block rather than from a length check both would pass. 576 checks, 0
  failures.
- **Simulation mode.** Turning the standard `SIMULATION` switch on before connecting connects the
  driver to a simulated controller with no port opened; `SIMULATED_GENERATION` on the Options tab picks
  which generation it imitates.

## Housekeeping

- GPL-2.0-or-later, with a licence header and named maintainer in every source file.
- Formatted with the repository's own `.clang-format`; the tree is clean under it.
- udev rule installs to `${UDEVRULES_INSTALL_DIR}` behind `INDI_INSTALL_UDEV_RULES`.
- The package builds with `fakeroot debian/rules binary` and passes lintian with one warning,
  `appstream-metadata-missing-modalias-provide`, deliberately not satisfied: the controller reports
  ST's generic `0483:5740`, so claiming that identifier would claim every stock-USB-stack STM32 device
  on the bus.
- Three commits: the driver, the packaging, and the top-level CMakeLists registration.


knro:
Thank you! Please remove the documentation from this PR add the driver documentation to [INDI Documentation Repository](https://github.com/indilib/drivers-docs) as it is required for all new drivers.

Regarding the udev rule, it appears to be STM32 CDC so it could be used for many devices and not just ScopeLink, am I correct? You already noted this in your submission above. So you need the symlink even though it could be pointing to different device? anything that could pin this down to your controller specifically?

tothbastrolabs:
Thank you — both done, branch updated.

**Documentation.** `doc/` is gone from the driver. The page and its eleven screenshots are submitted to
the documentation repository as indilib/drivers-docs#38, under
`src/content/docs/focusers/astrolabs/scopelink/`, and listed in dust caps, light boxes and auxiliaries as
well since the driver claims those interfaces too. The three manual pages moved to `indi-scopelink/man/`
rather than being removed, as the build installs them to `${CMAKE_INSTALL_MANDIR}/man1` — happy to drop
those too if you would rather the driver shipped none. `README.md`, `INSTALL.md` and `indi_scopelink.1`
now point at drivers.indilib.org.

**udev.** You are correct, and I have removed both symlinks. `0483:5740` is ST's generic Virtual COM Port
pair, and the firmware sets no product string either — a unit reports ST's default "STM32 Virtual ComPort
in FS Mode" — so nothing in the descriptors separates a ScopeLink from a Nucleo board or a home-built
focuser. `/dev/scopelink-<serial>` was no better than `/dev/scopelink` in that respect: it would have
given a foreign device a unique name rather than an ambiguous one. The rule now grants access and claims
no name at all.

Users lose nothing by it, because systemd already provides the per-unit name that symlink was rebuilding
by hand, from the same USB serial number:

```
/dev/serial/by-id/usb-STMicroelectronics_STM32_Virtual_ComPort_in_FS_Mode_205E35793630-if00
```

That is what the documentation now tells people to point an Ekos profile at.

To pin it down properly, we have requested our own USB product identifier from STMicroelectronics. Once
it is assigned the firmware will report it and the rule becomes an exact VID:PID match, symlink included;
I will submit that as a follow-up. Until then the driver's identify-at-connect check is the safety net —
it refuses anything that does not answer as a ScopeLink rather than half-driving an unknown device. I did
consider a `PROGRAM=` rule that probes the port at plug-in time, and rejected it: writing our protocol
into somebody else's controller is not a udev rule's business.

Rechecked on the updated branch: 0 compiler warnings, tests pass, `cmake --install` places all three
manual pages from `man/`, `man --warnings` renders them without complaint, and `udevadm verify` passes
the rule.

tothbastrolabs:
One more, found by reading the earlier macOS run rather than waiting for it to fail again: that failure
was mine, and it is fixed in the same push.

The job built the driver fine and fell over in `Install INDI 3rd Party Drivers`:

```
CMake Error at indi-scopelink/cmake_install.cmake:74 (file):
  file cannot create directory: /usr/lib/udev/rules.d.  Maybe need administrative privileges.
```

The udev rule was installed unconditionally, so macOS — which has no udev, and where the workflow
installs unprivileged into `/usr/local` — tried to create a Linux path it will never read. The install is
now guarded the way `libtoupcam` and `libsvbony` guard theirs:

```cmake
if(INDI_INSTALL_UDEV_RULES AND UNIX AND NOT APPLE)
```

Checked both ways: on Linux the rule still installs to `/usr/lib/udev/rules.d` and the three manual pages
to `share/man/man1`; configured with `APPLE` set, the udev entry is gone from `cmake_install.cmake` while
everything else still installs. `INSTALL.md` now says as much — macOS needs no rule, the controller is
`/dev/cu.usbmodem<serial>` there.

The workflow runs on the current head are held at "action required", so nothing has re-run since the
first round; they should go green when you approve them.

indi-scopelink/CMakeLists.txt
```diff
@@ -0,0 +1,131 @@
...
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/indi_scopelink.xml DESTINATION ${INDI_DATA_DIR})
+
+# Without this rule the controller comes up owned by root on Linux. It goes under /usr/lib rather than
+# /etc because /etc/udev is where the administrator's own overrides live, and a package has no business
+# writing there. Linux only: macOS has no udev, and installing there means creating a directory the
+# system will never read, which is what libtoupcam and libsvbony avoid the same way.
+option(INDI_INSTALL_UDEV_RULES "Install UDEV rules" On)
+
+if(INDI_INSTALL_UDEV_RULES AND UNIX AND NOT APPLE)
+  set(UDEVRULES_INSTALL_DIR "/usr/lib/udev/rules.d" CACHE STRING "Base directory for udev rules")
```
knro: Instead of setting this explicitly, add a condition:
```
if (NOT DEFINED UDEVRULES_INSTALL_DIR)
   set (UDEVRULES_INSTALL_DIR "/usr/lib/udev/rules.d" CACHE STRING "Base directory for udev rules")
endif()
```

indi-scopelink/CMakeLists.txt
```diff
@@ -0,0 +1,131 @@
...
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+find_package(INDI 2.0.0 REQUIRED)
+
+include(GNUInstallDirs)
+
+# The repository's shared compiler settings - C++17, the warning set, the hardening flags. OPTIONAL
+# because it lives in indi-3rdparty's cmake_modules rather than here, and a standalone development tree
+# only has FindINDI.cmake alongside it.
+include(CMakeCommon OPTIONAL)
```
knro: Why not inclukde cmake_modules in your tree? or link it. This way it always uses the common 3rd party headers. You don't then need to declare CXX_STANDARD..etc as it is already done.

indi-scopelink/CMakeLists.txt
tothbastrolabs: Done, thank you — it now reads exactly as you wrote it, which is what `libqhy` does too:

```cmake
if(INDI_INSTALL_UDEV_RULES AND UNIX AND NOT APPLE)
  if(NOT DEFINED UDEVRULES_INSTALL_DIR)
    set(UDEVRULES_INSTALL_DIR "/usr/lib/udev/rules.d" CACHE STRING "Base directory for udev rules")
  endif()
  install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/udev/99-scopelink.rules DESTINATION ${UDEVRULES_INSTALL_DIR})
endif()
```

Checked by configuring with `-DUDEVRULES_INSTALL_DIR=/tmp/somewhere/else`: `cmake_install.cmake` then names
that path and `/usr/lib/udev/rules.d` appears nowhere in it, where before the `set()` overwrote what the
packager asked for.

indi-scopelink/CMakeLists.txt
tothbastrolabs: You are right, and the `OPTIONAL` was the cause of the rest of it — because `CMakeCommon` might not be
found, the file restated settings it already makes. It is now a plain `include(CMakeCommon)`, the way the
other drivers here write it, and `CMAKE_CXX_STANDARD`, `CMAKE_CXX_STANDARD_REQUIRED`, `CMAKE_CXX_EXTENSIONS`
and the per-target `POSITION_INDEPENDENT_CODE` are all gone with it.

Nothing needs to move into the driver directory for that: inside a checkout the driver already sits beside
`cmake_modules`, which is what the first `CMAKE_MODULE_PATH` entry points at. The second entry stays for
building the driver on its own during development, where I keep a copy of the same modules next to it —
which is your "include cmake_modules in your tree", just on my side rather than in the submission.

Verified in tree after the change: the flags now come from the shared module and reach the compiler
(`-std=gnu++17 -Wall -Wextra -fPIC -fstack-protector-all -D_FORTIFY_SOURCE=2`), the build is still free of
warnings, the tests pass, `cmake --install` places the same files as before, and a configure with `APPLE`
set still drops only the udev entry.

One unrelated fix rode along in the same push: `README.md` and `indi_scopelink.1` linked to
`drivers.indilib.org/focusers/astrolabs/scopelink`, which the site answers with 403 — pages are served at
`.../scopelink/scopelink`. Both links now use the working form. Thank you for merging
indilib/drivers-docs#38.

PR #1373, opened 2026-08-20
Update based on Ferran Casarramona opened issue

s-pouya-sh:
Details:
https://indilib.org/forum/stellarmate/16009-sbig-on-stellarmate-os-2-3-solution.html#107807


knro:
This will not work on older systems. You need to fix this like how we do it for QHY rules.

Issue #1238, opened 2026-01-03
indi_toupcam_ccd offset problems (toupcam / toupbase), ToupTek ATR2600C (maybe others too?)


imaluschi:
I had some issues with Flats - overcorrections / undercorrections, so I started Investigations and taking new Flat-Darks, also Bias frames (for rare cases).
Following behaviour occured with Flat-Darks earlier too, but I was not sure if it has something to do with short exposure times for Flat-Darks (depending on the used filter down to sub 1 sec, at the moment I don't have a dimmer):
Taking Bias frames via Indi and KStars, the sequences were recorded directly one after the other. 
Exposure 0,0001 sec (the lowest I can set), "Low Noise mode", "High Conversion Gain", Gain 100
Set Offset to 200 (as used previously)
Taking Frames with Offset 200
Change Offset to 100
Taking Frames with Offset 100

**Describe the bug**

Some of the analyzed pictures showed different values in mean and median values.
* Bias Frames with Offset 100:
         mean and median around 103 (seems plausible)
         0,00197 % clipped pixels
         Histogram shows curve shifted to the right

![Image](https://github.com/user-attachments/assets/d322f5d5-b2c0-4222-9034-3ba0a98baf57)

* Bias Frames with Offset 200: 
         mean and median around 4-5 (makes no sense)
         20,8 % clipped pixels - every 5th pixel!!!
         The histogram shows that the curve is on the far left and cut off.

![Image](https://github.com/user-attachments/assets/f1830a7d-e417-41f1-9841-362728e9a2b1)

![Image](https://github.com/user-attachments/assets/6d48bfbc-d47b-4488-8ada-4ec5774833d1)

This problem occurred with other exposure times too, even shots with an exposure time of 150 seconds with Offset 100 were affected – the previously taken images with an exposure time of 120 seconds are just as good as the subsequent shots with an exposure time of 180 seconds.
Both LCG and HCG are affected.

![Image](https://github.com/user-attachments/assets/de388633-e7e9-473e-93ef-311e1875ad73)

It appears that the offset is not set arbitrarily, even though the offset value in the FITS header shows the value specified in the KStars Capture settings.

Important to know: all frames in a sequence have the same mean/median values, so either all or none frames are affected.

**To Reproduce**
Exact steps to reproduce the behavior.
1. Indi Server via Indi Web Manager App
2. Start KStars, Connect to Indi Web
[...]

PR #1374, opened 2026-08-20
Fix toupbase restore conversion gain taillight

### LINKED ISSUE #1319: indi_toupcam_ccd doesn't use saved offset
I have saved offset = 2000 for my ATR2600M. I did dark shot and looked at average of result image. Average is 254 but for offset = 2000 it have to be about ~2000. 
Then I changed offset to 2000 -> 2001 and back 2001 -> 2000, took shot and saw that average of image is about 2000 now.
So I have to change 2000 -> 2001 -> 2000 to make offset work correctly

INDI lib. v 2.2.0, Ubuntu, ARM64

### LINKED ISSUE #1361: Restore of toupcam settings broken after 5969bd9e
**Describe the bug**
Restore of conversion gain, taillight, all camera settings that worked before on Toupcam cameras now are not working again. Settings in the control panel, and the ones applied to the camera are not in sync. This will invalidate dark and bias libraries as subs will be taken with a different conversion gain if one does not set everything up manually on each and every indi startup.

**To Reproduce**
Start indi, change settings (taillight off, HDR as conversion gain on the Touptek 585), save settings, stop indi then start indi again.
Settings are loaded, and displayed correctly on the control panel, but camera shows a different tale: tail light is on, conversion gain is set to some unknown value.

**Expected behavior**
Saved values are restored, and applied to the camera on startup.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Not OS dependent
 - Version 2.2.1+

cfuture81:
## Summary
After commit 5969bd9e, several saved ToupTek settings are not applied to the
camera on connect unless a client replays the config. This makes the driver
apply them itself, so darks/bias and lights use the configured settings.

### 1. Conversion gain & tail light (Fixes #1361)
5969bd9e made setupParams() sync the INDI switch *from* the camera instead of
pushing the saved value, so with no config-load the camera keeps its power-on
default while the panel shows the wrong state. The tail-light readback also
inverted the state (option value 1=on written into switch index 1 = INDI_DISABLED),
displaying an enabled light as OFF. Fix: push the saved value to the camera in
setupParams(); keep GAIN_LOW = ISS_ON so the ISR_1OFMANY switch has a valid on-state.

### 2. Offset, low noise, high full well (Fixes #1319, Related #1238)
These reached the camera only via the client's config replay. On a fresh connect
the saved offset is ignored until the user nudges it (2000→2001→2000, #1319); left
at 0 it clips darks/bias (#1238). Fix: load() them in initProperties() and apply
them in setupParams(), offset LAST. IULoadConfigNumber() doesn't clamp, so large
offsets (2000) are preserved.

Scope note: the hypothesised #1238 trigger (a sensor-mode change resetting the
black level) could NOT be reproduced on an ATR2600C with the current SDK; the
reproducible failure is the offset not being applied on connect.

## Hardware validation (ATR2600C + GPM174M, no client config-load)
| Setting | Current driver | This PR |
|---|---|---|
| Conversion gain (saved LOW, default HIGH) | ran HIGH (std 5.66) | LOW (std 4.76) |
| Tail light (saved OFF, default ON) | LED ON, panel showed OFF | LED OFF, panel OFF |
| Offset (saved, cold boot) | mean ~5, 35% clipped | configured offset, 0% clipped |
| Low noise / high full well | not restored | restored |

CG verified via read noise, tail light via the camera LED, offset via bias-frame
statistics after a true 12V power cycle.

### DIFF
--- a/indi-toupbase/indi_toupbase.cpp
+++ b/indi-toupbase/indi_toupbase.cpp
@@ -156,6 +156,10 @@ bool ToupBase::initProperties()
     ///////////////////////////////////////////////////////////////////////////////////
     m_OffsetNP[0].fill("OFFSET", "Value", "%.f", 0, 255, 1, 0);
     m_OffsetNP.fill(getDeviceName(), "CCD_OFFSET", "Offset", CONTROL_TAB, IP_RW, 60, IPS_IDLE);
+    // Load the saved offset so it can be applied to the camera in setupParams(). The camera
+    // does not persist the black level (get_Option returns the default), so it must be
+    // restored from config and re-applied after any sensor-mode change that resets it.
+    m_OffsetNP.load();
 
     ///////////////////////////////////////////////////////////////////////////////////
     // R/G/B/Y Level Range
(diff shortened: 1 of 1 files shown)


knro:
Thank you, this has been going on for a while with various fixes. I hope this PR settles the matter!

PR #1377, opened 2026-08-25
indi-aagcloudwatcher-ng: persist Options and Rain Heater settings (#1365)

### LINKED ISSUE #1365: indi_aagcloudwatcher_ng: Sky Temp. Correction K1–K5 and other options not persisted across restarts
**Describe the bug**
In the `indi_aagcloudwatcher_ng` driver (AAG Cloud Watcher NG), the
"Sky Temp. Correction Parameters" (K1..K5, property `skyCorrection`) are
not persisted. After restarting the indiserver they always revert to the
skeleton defaults (33 / 0 / 4 / 100 / 100). Configuration → Save does not
include them: the property is neither written to
`~/.indi/AAG Cloud Watcher NG_config.xml` nor loaded at startup.

The same applies to `sqmLimit`, `anemometerType`, `heaterParameters`,
`heaterPIDParameters` and `heatingAlgorithm`.

**To Reproduce**
1. Start the driver and connect to the device.
2. Options tab → change K1 (e.g. 33 → 38) → Set.
3. Options tab → Configuration → Save.
4. Restart the indiserver, reconnect.
5. K1 is back to 33. The saved config file contains no `skyCorrection`
   vector.

**Expected behavior**
After Configuration → Save, the values should be written to the config file
and restored automatically when the driver starts (the usual INDI pattern:
`saveConfigItems()` override + `loadConfig(true, ...)` calls).

**Root cause (from current master)**
- `indi_aagcloudwatcher_ng_sk.xml` defines `skyCorrection` with hard-coded
  defaults; the corrected sky temperature is computed driver-side in
  `indi_aagcloudwatcher_ng.cpp` (`sendData()`) from these in-memory values.
- The driver does not override `saveConfigItems()`
  (see `indi_aagcloudwatcher_ng.h`), so only connection/debug/polling
  settings are saved.
- The driver itself never calls `loadConfig(true, "skyCorrection")`. Restoration 
therefore depends on the client pressing Load (Ekos/KStars does this automatically 
on connect - `kstars/ekos/manager.cpp`) - but since Save never writes the property, 
there is nothing in the file to restore. A hand-edited `skyCorrection` vector in the 
config file is picked up by Ekos' auto-load, which is my current workaround. It is 
lost again on the next Save.

**Proposed fix**

    bool AAGCloudWatcher::saveConfigItems(FILE *fp)
    {
        INDI::Weather::saveConfigItems(fp);
        getNumber("skyCorrection").save(fp);
        getNumber("sqmLimit").save(fp);
        getSwitch("anemometerType").save(fp);
        getNumber("heaterParameters").save(fp);
        getNumber("heaterPIDParameters").save(fp);
        getSwitch("heatingAlgorithm").save(fp);
        return true;
    }

    // in initProperties(), after buildSkeleton(...):
    loadConfig(true, "skyCorrection");
    loadConfig(true, "sqmLimit");
    loadConfig(true, "anemometerType");
    loadConfig(true, "heaterParameters");
    loadConfig(true, "heaterPIDParameters");
    loadConfig(true, "heatingAlgorithm");

**Side note / question**
The CloudWatcher hardware appears to store its own condition set onboard:
with the Windows AAG software, K-values/limits persist and the unit drives
its relay autonomously without a PC. The published RS232 protocol
(https://lunaticoastro.com/aagcw/TechInfo/Rs232_Comms_v100.pdf) documents
no write-parameters command, so the driver re-implements the correction in
software. If Lunatico can document the device-side storage, syncing with it
could be the better long-term fix.

**Screenshots**
Image

**Desktop:**
 - OS: Debian 13, KStars installed via Flatpak (stable)
 - kstars v3.8.3
 - CloudWatcher firmware v5.89

**Log Files**
No errors appear in the logs — the values are simply never saved/loaded
(verifiable in source and via the missing `skyCorrection` entry in the saved
config XML).

cfuture81:
## Summary

Fixes #1365. The AAG Cloud Watcher NG driver never persisted its user-configurable settings, so the Sky Temp. Correction parameters (K1..K5), SQM limit, anemometer type, rain-heater parameters, heating algorithm and heater PID parameters reverted to the skeleton defaults on every restart.

This adds a `saveConfigItems()` override for the six affected skeleton properties and restores them from an `ISGetProperties()` override. `deviceSwitch` is deliberately **not** persisted — it drives the unit's relay and must not be re-asserted automatically at startup.

# Why the load happens in `ISGetProperties()`, not `initProperties()`

The placement (`loadConfig()` in `initProperties()`) looks correct but silently does nothing. `loadConfig()` → `IUReadConfig()` → `dispatch()`, and `dispatch()` rejects any value for a property that is not yet defined:

```c
if (rosc_find(name, dev) == NULL) {
    snprintf(msg, MAXRBUF, "Property %s is not defined in %s.", name, dev);
    return -1;
}
```

`DefaultDevice::ISGetProperties()` runs `initProperties()` *before* its define loop, so at that point nothing is registered. And `IUReadConfig()` discards the `dispatch()` return value, so `loadConfig()` still reports success. Loading from `ISGetProperties()` after the base class has defined the (dynamic/skeleton) properties is the established pattern — same as `joystick.cpp`.

## Handler fixes required by the load path

A config load dispatches through `ISNewSwitch()`/`ISNewNumber()` with the **whole** vector, whereas a GUI client sends only the widget it changed. Three spots assumed the client-style call:

- **`heatingAlgorithm`** derived `usePIDforHeating` from `names[0]`, which on a config load is not necessarily the selected switch. Result: the Options tab shows PID selected while the driver keeps running the original algorithm. Now derived from the resulting switch state.
- **`anemometerType`** passed a hard-coded `n` of 2 to `update()`, reading past the end of `states[]`/`names[]` when a client sends a single switch.
- **`skyCorrection`** clamped `values[0..4]` regardless of `n`.

## Testing

No hardware required — `saveConfigItems()`/`loadConfig()` run in the property layer, before and independently of `Handshake()`. Verified end-to-end against a plain `indiserver` using `indi_setprop`/`indi_getprop`:

- All six vectors are written on Save and restored after an `indiserver` restart.
- `usePIDforHeating` is correctly `true` after loading `pid=On` (the naive patch leaves it `false`).
- The unpatched binary reverts all six to defaults and writes none of them to the config file.
- Fresh install with no config → skeleton defaults; missing skeleton file → no crash, no junk written.

**Not verified without hardware:** the anemometer wind-speed scaling side effect. `setAnemometerType()` only affects scaling applied during real sensor reads; the switch state restores and the handler runs, but the resulting km/h value wasn't exercised. A sanity check by someone with grey-cup hardware would be appreciated.

---

## Update — `sqmLimit` broadcast fix (2nd commit)

Follow-up to GUI verification by @zero-drift-ops. Of the six persisted vectors, `sqmLimit` was the only one without an `ISNewNumber` handler, so it fell through to `DefaultDevice::ISNewNumber`. That path calls `PropertyNumber::update()` (which writes the value into storage) but, with no registered update callback, returns without ever calling `apply()`. `apply()` is what emits `setNumber` to clients, so on a config load the value was applied internally but never broadcast: the client attached across the define→load window (the Ekos/KStars GUI) kept the skeleton default (19.60) while the driver used the loaded value. The same missing broadcast also meant a `sqmLimit` change from one client was not pushed to other connected clients. `indi_setprop`/`indi_getprop` masked both, since a fresh client re-reads the property from the already-updated storage — which is why the round-trip test above didn't catch it.

The 2nd commit adds a `sqmLimit` handler (`update`/`apply`) mirroring the existing `heaterPIDParameters`/`skyCorrection` branches; `update()` already clamps to the skeleton min/max. No change to the save/load logic. Verified: builds clean, round-trip still passes, and (per @zero-drift-ops) set SQM=20 → Save → full Ekos/KStars restart now shows 20 (was reverting to 19.60), and a value set by one client is now broadcast to a second connected client.

### DIFF
--- a/indi-aagcloudwatcher-ng/indi_aagcloudwatcher_ng.cpp
+++ b/indi-aagcloudwatcher-ng/indi_aagcloudwatcher_ng.cpp
@@ -140,6 +140,51 @@ bool AAGCloudWatcher::initProperties()
     return true;
 }
 
+/**********************************************************************
+ ** Restore the saved settings.
+ **
+ ** This must happen after the base class has defined the properties:
+ ** the driver-side dispatcher rejects a value for a property that has not
+ ** been defined yet, so calling loadConfig() from initProperties() silently
+ ** does nothing.
+ **
+ ** Loading goes through the regular ISNew*() handlers, so driver state derived
+ ** from these properties (anemometer type, heating algorithm) is set up as well.
+ ** Doing it here instead of relying on the client pressing "Load" keeps the
+ ** settings working under a plain indiserver too.
+ **********************************************************************/
+void AAGCloudWatcher::ISGetProperties(const char *dev)
+{
+    INDI::Weather::ISGetProperties(dev);
+
+    if (!m_ConfigLoaded)
+    {
+        m_ConfigLoaded = true;
+        for (const auto &property : CONFIGURABLE_PROPERTIES)
+            loadConfig(true, property);
+    }
+}
+
+/**********************************************************************
+ ** Persist the user-configurable settings.
+ **********************************************************************/
+bool AAGCloudWatcher::saveConfigItems(FILE *fp)
+{
+    INDI::Weather::saveConfigItems(fp);
+
+    // buildSkeleton() may have failed (missing indi_aagcloudwatcher_ng_sk.xml), in
+    // which case these properties do not exist. save() on an invalid property is a
+    // no-op, but check explicitly so the intent is clear.
+    for (const auto &name : CONFIGURABLE_PROPERTIES)
+    {
+        auto property = getProperty(name);
+        if (property.isValid())
+            property.save(fp);
+    }
+
+    return true;
+}
+
 IPState AAGCloudWatcher::updateWeather()
 {
     // in case elevation updated as GPS gets a better fix
@@ -274,7 +319,8 @@ bool AAGCloudWatcher::ISNewNumber(const char *dev, const char *name, double valu
 
             if (nvp.isNameMatch("skyCorrection"))
             {
-                for (int i = 0; i < 5; i++)
+                // Bound by n, not by the vector size: a client may send a subset.
+                for (int i = 0; i < n; i++)
                 {
                     if (values[i] < -999)
                     {
@@ -310,11 +356,18 @@ bool AAGCloudWatcher::ISNewSwitch(const char *dev, const char *name, ISState *st
 
             if (svp.isNameMatch("heatingAlgorithm"))
             {
-                LOGF_INFO("Changing heating algorithm to %s\n", names[0]);
-                usePIDforHeating = (strcmp(names[0], "pid") == 0);
                 svp.update(states, names, n);
                 svp.setState(IPS_OK);
                 svp.apply();
+
+                // Derive the flag from the resulting switch state rather than from
+                // names[0]. A GUI client sends only the switch it turned on, but a
+                // configuration load sends the whole vector, in which case names[0]
+                // is not necessarily the selected one.
+                auto pid = svp.findWidgetByName("pid");
+                usePIDforHeating = (pid != nullptr && pid->getState() == ISS_ON);
+                LOGF_INFO("Changing heating algorithm to %s", usePIDforHeating ? "pid" : "original");
+
                 return true;
             }
 
@@ -394,12 +447,14 @@ bool AAGCloudWatcher::ISNewSwitch(const char *dev, const char *name, ISState *st
 
             if (svp.isNameMatch("anemometerType"))
             {
-                svp.update(states, names, 2);
+                // Use n, not a hard-coded 2: a client may send only the switch it
+                // turned on, and reading past the end of states[]/names[] is UB.
+                svp.update(states, names, n);
                 svp.setState(IPS_OK);
                 svp.apply();
 
                 auto sp = svp.findWidgetByName("BLACK");
-                if (sp->getState() == ISS_ON)
+                if (sp != nullptr && sp->getState() == ISS_ON)
                 {
                     cwc->setAnemometerType(BLACK);
                 }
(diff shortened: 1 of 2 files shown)


zero-drift-ops:
Found one more gap during GUI verification of #1377: `sqmLimit` is the only
one of the six vectors without an `ISNewNumber` handler in the driver.

A config load dispatches through `ISNewNumber`/`ISNewSwitch`.
The other five vectors have driver handlers that end in `nvp.apply()` /
`svp.apply()`, which broadcasts the restored value to connected clients.
`sqmLimit` falls through to `DefaultDevice::ISNewNumber`, which calls
`property.update(values, names, n)`. The value is applied internally but
never sent (`apply()` is what broadcasts; `update()` only emits an internal
callback). So a client that received the pre-load definitions keeps the
skeleton default (19.6) after restart, while every other vector updates.
`indi_setprop`/`indi_getprop` testing masks this: `indi_getprop` re-requests
the property after the load completed, so it always reports the loaded value.

Fix (carried locally until it lands):

```diff
@@ -317,6 +317,19 @@
                 return true;
             }
 
+            // Broadcast the loaded value to clients. Without a driver handler,
+            // sqmLimit falls through to the generic DefaultDevice::ISNewNumber
+            // path, which applies values internally but never sends (apply) them,
+            // so a config load would silently leave clients on the default.
+            if (nvp.isNameMatch("sqmLimit"))
+            {
+                nvp.update(values, names, n);
+                nvp.setState(IPS_OK);
+                nvp.apply();
+
+                return true;
+            }
+
             if (nvp.isNameMatch("skyCorrection"))
             {
                 // Bound by n, not by the vector size: a client may send a subset.
```

Verified on my imaging PC. With the patched driver, set SQM Limit to 20,
Save, full Ekos/KStars restart → shows 20 (was reverting to 19.60 before).
Also verified at the property layer that a value set by one client is now
broadcast to a second connected client (was silent before).

Not verified: anemometer wind-speed scaling side effect (unit has the BLACK
cups).

cfuture81:
@zero-drift-ops 
Good catch & confirmed. 
sqmLimit was the only one of the six without an ISNewNumber handler, so on load it fell through to DefaultDevice::ISNewNumber, which update()s the value but doesn't apply() it. 
Added the handler exactly as you suggested (matches the existing heaterPIDParameters/skyCorrection shape); builds clean and the round-trip still passes. Pushed as [commit]. Appreciate the GUI-side verification.

cfuture81:
2nd commit added

PR #1384, opened 2026-09-02
QHY: persist overscan area setting in driver configuration

etlam262:
## What

Persist the QHY `OVERSCAN_MODE` switch (`OverscanAreaSP`) in the driver
configuration by saving it in `QHYCCD::saveConfigItems()`.

## Why

`OVERSCAN_MODE` is user-settable, but it was never written to the config
file. As a result the switch always reverted to "Ignore" on driver
startup, regardless of what the user had saved.

## How

The switch is saved alongside the other optional properties, guarded by
`HasOverscanArea`, consistent with how `AMPGlowSP`, `GPSControlSP` and the
other feature-gated properties are handled.

It is deliberately written **after** `INDI::CCD::saveConfigItems()`. On
config load INDI replays properties in file order, so emitting the switch
after the base call means it is applied after `CCD_FRAME`, letting the
existing `ISNewSwitch` handler rebuild the sensor ROI last.

## Reviewer note — interaction with a saved custom ROI

- In the default **"Ignore"** mode a saved sub-frame is preserved: the
  `ISNewSwitch` handler early-returns when the restored mode equals the
  current one, and never touches `CCD_FRAME`.
- If the saved mode is **"Include"**, applying it on load calls
  `UpdateCCDFrame()` with the full sensor area, which overrides the
  just-restored sub-frame. This is inherent to how "Include" redefines the
  frame as the whole sensor — identical to toggling the switch
  interactively — and only affects that mode.

Happy to move the ROI-preserving behaviour into the handler instead if
maintainers prefer, but that is a larger behaviour change than this
persistence fix.

## Testing

Tested with a QHY268C on arm64 (indi-qhy 2.9, INDI 2.2.4): the overscan
selection now survives a driver restart.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


knro:
Thank you, looks good to me!

PR #1385, opened 2026-09-08
Nicknames for ASI filter wheels

sterne-jaeger:
With multiple ZWO EFW filter wheels connected, indi_asi_wheel names them "ZWO EFW" / "ZWO EFW 1" by USB enumeration order, not physical identity — so which wheel gets which name isn't stable across restarts, breaking per-wheel filter configs.

**Solution**
Wire ASIWHEEL up to INDI core's existing nickname API (addNicknameControl/setDeviceNicknameFromId/saveNicknameId), the same mechanism the EAF focuser driver already uses. The hotplug handler now fetches each wheel's real hardware serial (EFWGetSerialNumber) and uses it to resolve/persist a nickname, independent of enumeration order. Serial-formatting logic was factored into a new asi_serial_helpers.h.

**Hints for the rotator follow-up**
ASICAA's Loader::load() already computes a serial number (CAAGetSerialNumber) but discards it. Applying the same fix there just means: thread that existing serial into the ASICAA constructor, add the same setDeviceNicknameFromId/addNicknameControl/nicknameSet trio, and format it via Helpers::toHexString instead of the rotator's own inline sprintf — no new scaffolding needed.


knro:
What happens to filter wheels without a serial number?

sterne-jaeger:
> What happens to filter wheels without a serial number?

If no serial number is present, the old logic is used - see indi-asi/asi_wheel.cpp::70.

sterne-jaeger:
Tested under real conditions, works fine, could be merged!

PR #1391, opened 2026-09-22
Add direct V4L2 support to QHY CCD driver

lft0323:
Add support for QHY V4L2 camera (QHY26800A)

This part of the code mainly uses the v4l2 driver framework in Linux for identification and control. Device identification is achieved by recognizing the video node corresponding to the v4l2 camera through the /dev/video node.

After connection, query the device capabilities to update the parameter range and set the camera parameters through the /dev/v4l-subdevx node.

Tested with prototype machine.

liu fu tang 
lft@qhyccd.com


knro:
Thanks, but how is this driver different from the base V4L2 driver?

lft0323:
These modifications are intended to distinguish the types of devices and customize the vendor name. As far as I know, the standard V4L2 driver does not support this functionality. 

This part of the function only runs on the Linux system inside the camera. On ordinary Linux computers, the traditional SDK connection will still be used.

knro:
So this driver is only internal to the camera itself? Can someone install this driver and connect to the camera directly? or you intend to have an INDI server running in side the camera?

lft0323:
Yes, the current mode is to run an indi server inside the camera. This driver is only used within the camera. 

In the future, Raspberry Pi with camera modules can also be used in conjunction with specific driver programs.

knro:
Ok a couple of issues:
1. Why there is no separate executable for this driver?
2. If this is intended *only* for internal camera use, where users cannot use it to connect to the camera itself, then it might not be useful to include here as you can simply just added it to the Linux system inside the camera. That is, any drivers not directly usable by users should not be included in the public repo.

lft0323:
I think you're right. We need to have a discussion within the team.

